PR #303, opened 2015-01-12
Ticket/292 - part 1

melinath:
Resolves half of #292 - display of whether an order has pending payments in the organizer model tables.

PR #304, opened 2015-01-14
Ticket/203

melinath:
This includes pull request #303 and should probably be merged separately and afterwards.

This pull request also fixes some issues with dwolla & stripe payment.

It would be nice to have a go-through of the UI as part of this pull request to add appropriate notices in places as to whether an event is a test event or a live event. For example, would be nice to have something on the event settings page. It would also potentially be good to only warn about insecure connections on live event pages, and only provide links to test cards on testing event pages. Or something like that.


harrislapiroff:
Got this migration error:

```
django.db.utils.IntegrityError: NOT NULL constraint failed: brambling_person__new.stripe_test_customer_id
```

I think we need a default value on `stripe_test_customer_id`?

melinath:
@harrislapiroff so... is there anything left on this?

harrislapiroff:
I haven't fully tested this. I'll work on that now.

harrislapiroff:
I did some rudimentary testing. I don't feel 100% confident about my coverage, but I can't think of how to be more thorough. I got some weird errors when refunding:

screenshot 2015-01-20 17 43 53

Though maybe refunds should be disallowed in test mode altogether? Although it would be nice to be able to test our refund code, eh...?

Otherwise, that and the bugs I fixed are all I found. Bouncing back to you @melinath.

melinath:
Did you get all those errors at the same time? The first two are probably from Dwolla - their test APIs are kind of terrible - essentially, you can only refund if your dwolla account has money in it. Not sure why you got two, unless the messages got stored up somehow...

harrislapiroff:
I might have hit refund a couple times and they didn't display till I visited a page that had messages in the template. (Also a bug?)

harrislapiroff:
I might also have picked an order that had multiple payments (with different methods) on it

PR #307, opened 2015-01-16
Ticket/289

nmorduch:
YUP

### DIFF
--- a/brambling/templates/brambling/event/order/_check_payment_info.html
+++ b/brambling/templates/brambling/event/order/_check_payment_info.html
@@ -1,9 +1,22 @@
-<p>Checks should be made out to <strong>{{ event.check_payable_to }}</strong> and include your order code (<strong>{{ order.code }}</strong>) in the memo line. Mail them to:</p>
-<p>
-	{{ event.check_recipient }}</br />
-	{{ event.check_address }}</br />
-	{% if event.check_address_2 %}{{ event.check_address_2 }}<br />{% endif %}
-	{{ event.check_city }}, {{ event.check_state_or_province }} {{ event.check_zip }}
-	{{ event.check_country }}
-</p>
-<p>Checks must be postmarked no later than {{ event.check_postmark_cutoff|date:"F jS, Y" }}. If you do not complete payment by that date, the event organizer reserves the right to revoke your passes.</p>
+{% load zenaida %}
+
+{% td_to_dict event.check_postmark_cutoff|td_timeuntil as time_until_check_cutoff %}
+
+{% if time_until_check_cutoff.days > 0 %}
+
+	<p>Checks should be made out to <strong>{{ event.check_payable_to }}</strong> and include your order code (<strong>{{ order.code }}</strong>) in the memo line. Mail them to:</p>
+	<p>
+		{{ event.check_recipient }}</br />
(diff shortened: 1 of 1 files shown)


brambling/templates/brambling/event/order/_check_payment_info.html
```diff
@@ -1,9 +1,22 @@
...
-	{% if event.check_address_2 %}{{ event.check_address_2 }}<br />{% endif %}
-	{{ event.check_city }}, {{ event.check_state_or_province }} {{ event.check_zip }}
-	{{ event.check_country }}
-</p>
-<p>Checks must be postmarked no later than {{ event.check_postmark_cutoff|date:"F jS, Y" }}. If you do not complete payment by that date, the event organizer reserves the right to revoke your passes.</p>
+{% load zenaida %}
+
+{% td_to_dict event.check_postmark_cutoff|td_timeuntil as time_until_check_cutoff %}
+
+{% if time_until_check_cutoff.days > 0 %}
```
melinath: Doesn't this mean that the address won't be available *on* the cutoff date?

brambling/templates/brambling/event/order/_check_payment_info.html
```diff
@@ -1,9 +1,34 @@
...
+	<p>
+		{{ event.check_recipient }}</br />
+		{{ event.check_address }}</br />
+		{% if event.check_address_2 %}{{ event.check_address_2 }}<br />{% endif %}
+		{{ event.check_city }}, {{ event.check_state_or_province }} {{ event.check_zip }}
+		{{ event.check_country }}
+	</p>
+	<p>Checks must be postmarked no later than <strong>{{ event.check_postmark_cutoff|date:"F jS, Y" }}</strong>. If you do not complete payment by that date, the event organizer reserves the right to revoke your passes.</p>
+
+{% elif time_until_check_cutoff.days = 0 %}
```
melinath: should be `== 0`

nmorduch: fixed with ab30954

melinath:
I feel like the copy here is generally kind of terrifying. Which is to say, the first two sections say "If you do not complete payment by that date" when what they should say is "If you do not postmark your payment by that date" or something. And the last section should say something like "If you've already sent in your payment, great!" Right now it sounds like my pass can be revoked if payment hasn't actually been received by the postmark date.

melinath:
Is it even possible to choose a different payment method once you've chosen check?

nmorduch:
@melinath It's not, but it should be. I'll make a ticket.

harrislapiroff:
I know this is merged, but I actually think "If you've already sent in your payment, great!" is a false security! Someone who is paying by check still needs to be attentive to whether or not their check gets received and contact an organizer if it doesn't. This message seems to indicate that they're fine for payment when they very well may not be. They _should_ be worried if an organizer has not marked their check as received.

harrislapiroff:
I think people are generally smart enough to know that their check could not possibly be marked as received if they mailed it yesterday, but I don't think people are smart enough to check back if we tell them it's all good.

melinath:
That's a good point.

harrislapiroff:
Okay, I'll make a ticket.

PR #309, opened 2015-01-16
Ticket/291

nmorduch:


brambling/views/orders.py
```diff
@@ -318,10 +319,10 @@ def get_order(self):
 
     def create_order(self):
         person = self.request.user if self.request.user.is_authenticated() else None
-        code = get_random_string(8)
+        code = get_random_string(8,ORDER_CODE_ALLOWED_CHARS)
```
melinath: PEP 8: there needs to be a space after the comma here and below.

nmorduch: Fixed in b20ba4e

Issue #292, opened 2014-12-29
pending payments display in order summary


nmorduch:
display received payments & refunds in one list, with a subtotal, then below that add pending payments, with a total.

rather than just all payments in one list, with a column for whether they've been received.

melinath:
@nmorduch This seems like you'll probably need differently structured data. Does that seem accurate?

melinath:
i.e. it can't be fixed by just changing the templates.

nmorduch:
Yeah. And I think that's it's important for OJDF, because I also want to display in the list of all orders how much each order has in outstanding payments.

nmorduch:
I think that in all cases, promising to send a check should not mean that you are marked as paid. You should be marked as pending payment (though with your order complete).
It's super confusing for the organizers otherwise; when everyone pays by check, the event summary says that none of the payments are outstanding, rather than giving a useful overview of how much money has actually been received.

melinath:
+1

melinath:
As part of this, we should also distinguish between cash in hand and pending payments in the event summary data.

melinath:
Partially resolved with merge of #303

PR #315, opened 2015-01-21
Updated event summary to include refund finances and differentiate between pending and completed payments.

melinath:
@nmorduch I think this resolves the event summary page display for now - did you want to include the order summary improvements as part of this ticket / include changes there? The data should now all be available. If there's anything missing let me know.

### DIFF
--- a/brambling/templates/brambling/event/organizer/summary.html
+++ b/brambling/templates/brambling/event/organizer/summary.html
@@ -20,20 +20,24 @@
 				</div>
 				<ul class='list-group'>
 					<li class='list-group-item'>
-						Gross sales
-						<strong class='pull-right text-info'>{{ gross_sales|format_money:event.currency }}</strong>
+						Payments received
+						<strong class='pull-right text-info'>{{ confirmed_payments|format_money:event.currency }}</strong>
+					</li>
+					<li class='list-group-item'>
+						Payments pending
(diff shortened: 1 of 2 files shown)


nmorduch:
Before I do this, does it work with the new internationalization?

melinath:
Yes, it should. Stripe charges use the event's currency, which was already changeable. The only "internationalization" that I did was actually displaying the correct country and currency on the event settings page.

melinath:
So this works as well as what we already had.

nmorduch:
Closes #292

nmorduch:
I did this using rebase in the command line and the commit went onto master but github thinks this branch is not merged. Did I do something wrong? I normally just merge using the button that says merge, but I didn't want the merge commit that comes with that.

melinath:
If you're merging from the command line, you do `git merge <branch> --no-ff` from the master branch. This makes a merge commit to show that a branch was merged. Rebase just sticks the branch on the end of the master branch as if it had always been there. If you then merge without --no-ff, the master fast-forwards.

Generally speaking, we want to have merge commits.

nmorduch:
But if we always have a new branch for every little thing, then that's a million merge commits. 

What should I do to remedy this now?

melinath:
Meh, just delete the branch.

PR #316, opened 2015-01-28
Order summary

nmorduch:
Makes the order summary page feel a little more receipt-like. You should look at it and see if it makes sense to you. It will no doubt need changes as we go into the multiple orders thing, but a lot of the styles will be useful.


nmorduch:
@harrislapiroff can you look at this too?

melinath:
I rebased this branch onto the latest master... it looks like there were some template variable name changes that made this version of the template no longer work correctly. But overall I do like it, I think. It'll need a bit of work to bring it up to date.

melinath:
This might actually be something broken due to a lack of internationalization support after all.

melinath:
Yeah, littleweaver/django-zenaida@d506bba5c0597fcdd292f8af3391517fc1e74b1f fixed the issues I was seeing. I think I like this more, but it does have less information, but that might be okay. I'd be interested in what @harrislapiroff thinks.

melinath:
It's a little weird to me that the cost is in a completely different place than the payments / refunds. I like the items being grouped on the side, but maybe they shouldn't list prices?

nmorduch:
How about this? When transactions become self-aware, we should:

- Remove prices from the Attendee in formation
- Make each transaction have a dropdown that says what things were bought for how much and for whom.

melinath:
Remaining dev work related to this: RefundView needs to check whether an order is fully or partially refunded and mark it appropriately. I think? Or should we just be annotating the Order model more aggressively with payments received and total refunds and not have status there at all?

Also, manual payments should probably mark any current cart as paid and attach the items to the payment. Manual payment page should also display currently-owed amount (and default the payment amount to that.)

nmorduch:
@melinath is this currently waiting on more development or do you want it merged?

nmorduch:
As far as I'm concerned this can be merged, but @harrislapiroff should be the one to review the python.

melinath:
@nmorduch I thought there was still more design work on this.

melinath:
That's why i assigned it to you, not for review, IIRC. Though I probably should have said that with words. :-p

nmorduch:
I think it should be merged and i can do additional work on a separate pull request.

harrislapiroff:
I think it would be nice to have payment information associated with each item (at minimum display a status indicator for "Unpaid", "Paid", or "Refunded", but this is definitely very nice. Merging.

Another thought: ultimately we might want separate views for "Your items" and "Your receipts"

PR #317, opened 2015-01-28
Closed #311 prepopulate check recipient

### LINKED ISSUE #311: Organizer Check Payment Form UI Improvements
* [ ] "Mail addressed to" field should be prepopulated with the value from the "Checks made out to" field.

nmorduch:


melinath:
Why not repurpose Django's prepopulate script? This works, but it seems like it would be more intuitive to have it be filled as you typed.

melinath:
Although I guess all you'd really have to do is hook to the keyup event on the first field instead of the focus event of the second one. Slash make sure the field being autofilled was empty to start with.

nmorduch:
I thought about having it go as you type, but that's actually kind of distracting to me, I think. Like, you only thought you were typing in this one box but suddenly there's something else happening in another box. This way, when something changes, it's in the box you're paying attention to right now. I find it less jarring. @harrislapiroff, intuition tie-breaker plz?

harrislapiroff:
I'm w/ Stephen here. I think it's preferable to prepopulate as you type. It's fairly common to edit something on one part of a page and have another part of a page update in real-time (whether that's django's prepopulate script for slugs or a live preview of some sort of input). I don't think it's excessively distracting. I think it's more of an unexpected surprise for things to happen without focus.

nmorduch:
k.

nmorduch:
Ok so I have 

```javascript
		$('#id_check_payable_to').focus(function() {
			$this = $(this), $recipient = $('#id_check_recipient');
			if ($recipient.val()=='') {
				$this.keyup(function(){
					$recipient.val($this.val());
				});
			}
		});
```

But then how do I kill the keyup when you unfocus?

The case I'm trying to avoid is:
- You edit the first input, it copies to the second, all is well and good
- You make an edit to the second input
- You realize something is wrong with this first input, so you go back to it, but the old keyup is still there so as you change it, it destroys your edits to the second input

melinath:
Here are a couple ways you could implement that:

```
        var $payable_to = $('#id_check_payable_to'),
            $recipient = $('#id_check_recipient'),
            recipient_val = '';
        $payable_to.on('keyup', function(e) {
            if ($recipient.val() === recipient_val) {
                recipient_val = $payable_to.val();
                $recipient.val(recipient_val);
            }
        });
```
```
        var $payable_to = $('#id_check_payable_to'),
            $recipient = $('#id_check_recipient'),
            do_the_thing = $recipient.val() === '';
        // This should *actually* use jQuery's one-time event binder
        $recipient.on('change', function(e) {
            do_the_thing = false;
        });
        $payable_to.on('keyup', function(e) {
            if (do_the_thing) {
                $recipient.val($payable_to.val());
            }
        });

```

harrislapiroff:
You might look at Django's prepopulate script and how it works. It uses a
data attribute on the field to keep track of whether or not it's been
modified. You might even copy it wholesale and just remove the URLify bit.

https://github.com/django/django/blob/master/django/contrib/admin/static/admin/js/prepopulate.js

harrislapiroff:
Although @melinath's code might be more compact.

PR #318, opened 2015-01-28
copy inputs between signup and login forms on splash

nmorduch:
Questions:

- Is it a bad idea to copy over the password?
- Can using the login/signup be even more similar and streamlined in the future?

I figured I'd do this while I still had it fresh from #311, since I've been meaning to work on this forever.

### DIFF
--- a/brambling/templates/brambling/splash.html
+++ b/brambling/templates/brambling/splash.html
@@ -18,9 +18,9 @@
 			</div>
 			<div class="col-sm-6">
 				<ul class="nav nav-pills masthead-pills text-jumbo" role="tablist">
-					<li class="active"><a href="#signup" class="masthead-link" role="tab" data-toggle="tab">Sign Up</a></li>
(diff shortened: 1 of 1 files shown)


melinath:
I think I would feel uncomfortable having my password copied. But maybe less tech savvy people wouldn't care?

nmorduch:
The idea is that it's not supposed to feel like it's being copied. It's supposed to feel like it was the same field. But I'm wondering if creating that illusion requires a real security risk.

melinath:
If it works, there's no security risk. And as long as we require them to fill out a second password field to make sure they match, it should be fine.

I am a little curious as to whether the copied password is actually copied correctly...

melinath:
If so, I'd say it's probably fine?

harrislapiroff:
I'm fairly sure this should copy the password correctly. I haven't tested it, but the only thing special about password fields versus text fields is that they mask the password in the UI. Javascript should still have access as usual.

harrislapiroff:
I will also say there's no security risk. The password is already exposed on the frontend. Manipulating it with javascript does not expose it any further.

nmorduch:
It worked correctly for me. I tested it, obviously :p

PR #319, opened 2015-01-28
closed #314 check payment info text

### LINKED ISSUE #314: Check text
The check text in the final `else` [here](https://github.com/littleweaver/django-brambling/blob/master/brambling/templates/brambling/event/order/_check_payment_info.html) should indicate that the organizer has not marked the check as received and that the attendee needs to make sure it's received. (Currently it says "If you've already sent in your payment, great!")

nmorduch:


melinath:
This include is used in the following contexts (as well as once on its own, in a place where the question of whether check payments are unreceived wouldn't make sense):

```
# brambling/event/order/summary.html
   40  			<p>Your passes are reserved, but the organizer has not yet received your payment by mail.</p>
   41: 			{% include "brambling/event/order/_check_payment_info.html" %}
   42  			<p>If you believe this is in error, please contact the event organizer directly.</p>
```

```
# brambling/mail/order_receipt_body.html
   11  	<p>Note: Your passes are reserved, but the organizer has not yet received your payment by mail.</p>
   12: 	{% include "brambling/event/order/_check_payment_info.html" %}
   13  	<p>If you believe this is in error, please contact the event organizer directly.</p>
```

The proposed change would end up making the phrasing in both places awkward, since it would repeat the fact that the payment hasn't been received.

Looking back, maybe #314 doesn't actually need to be fixed? Or maybe the fix needs to also touch other files?

nmorduch:
Mm yeah you're right. This is probably why I wrote it the way it was written originally. @harrislapiroff, thoughts?

harrislapiroff:
My primary concern for #314 which remains a concern with this pull request is that, "If you've already sent in your payment, great!" is misleading. There's no way for us to know that things are great. If you've already sent in your payment and your check gets lost in the mail, that's still not great! I understand that @melinath was concerned that the text prior to adding that sentence sounded too severe, but I don't think we should be reassuring people that things are fine when they are not necessarily.

harrislapiroff:
Basically, the situation, unfortunately, is that as long as the organizer has not marked your check as received, even if you have in fact mailed it, you may still be turned away from the event and we shouldn't imply otherwise.

harrislapiroff:
Honestly, I would just strike that sentence and let it go back to the more severe form, but if we wanted a softer form, maybe something like, "If it has been more than a week since you mailed your check, you should make sure that it was correctly addressed and contact the organizer." or something? But that will still require some reformulation to reduce awkwardness since some of the include locations also include "If you believe this is in error, please contact the event organizer directly."

nmorduch:
@harrislapiroff I think that at this point you'd be the best person to come up with the best wording. Gonna assign it to you.

nmorduch:
Except it's already assigned to you.

harrislapiroff:
Included in the new commits, the new text (from the event summary page) reads like so on the event summary page:

Before due date:

>  Your passes are reserved, but the organizer has not yet received your payment by mail.
> 
> Checks should be made out to Harris Lapiroff and include your order code (1j0eK27j) in the memo line. Mail them to:
> 
> Harris Lapiroff
> 123 Main St
> Anytown, ST 99999
> 
> Checks must be postmarked no later than February 12th, 2015. If your payment is not postmarked by that date, the event organizer reserves the right to revoke your passes.

> If you mailed your check more than a week ago or believe this is in error, please contact the event organizer directly.

On due date:

> Your passes are reserved, but the organizer has not yet received your payment by mail.
> 
> Checks should be made out to Harris Lapiroff and include your order code (1j0eK27j) in the memo line. Mail them to:
> 
> Harris Lapiroff
> 123 Main St
> Anytown, ST 99999
> 
> Checks must be postmarked no later than today. If your payment is not postmarked by today, the event organizer reserves the right to revoke your passes.
> 
> If you mailed your check more than a week ago or believe this is in error, please contact the event organizer directly.

After due date:

> Your passes are reserved, but the organizer has not yet received your payment by mail.
> 
> The organizer of this event has not yet received your check. Checks should be postmarked no later than January 8th, 2015. The event organizer reserves the right to revoke your passes if you have not sent in your payment.
> 
> If you mailed your check more than a week ago or believe this is in error, please contact the event organizer directly.

harrislapiroff:
Sending back to either @nmorduch or @melinath for review.

harrislapiroff:
Oh wait, that last one is repetitive still.

harrislapiroff:
Fixed.

melinath:
Very nice. The wording for the email doesn't change - but it doesn't really need to, since presumably they don't send a check before they register.

PR #323, opened 2015-02-05
Resolved #310

### LINKED ISSUE #310: Prepopulate.js does not work if Grappelli is installed
Because grappelli's version of prepopulate calls for `grp.jQuery` instead of `django.jQuery`

https://github.com/littleweaver/django-brambling/blob/master/brambling/templates/brambling/event/organizer/create.html#L16

harrislapiroff:


harrislapiroff:
@nmorduch or @melinath one of you should give this a quick look-over. It's just a one-liner.

nmorduch:
I don't quite understand what's going on here. @harrislapiroff can you explain to me, or @melinath can you review?

melinath:
So... it looks like this essentially overrides the versions of jQuery that django and grappelli use. But... it doesn't seem like this should work, looking at [grappelli.js](https://github.com/sehmaschine/django-grappelli/blob/2.6.3/grappelli/static/grappelli/js/grappelli.js). It looks like the grappelli functionality is initialized with grp.jQuery *in* that file. So changing the value of grp.jQuery afterwards shouldn't make a difference.

harrislapiroff:
Grappelli.js is not included in this page. (From the admin,) the page _only_ includes the prepopulate script, which uses either `django.jQuery` or `grp.jQuery`, depending on whether or not grappelli is installed. Without this line, `django` and `grp` is undefined and the script fails.

Test it out on an event creation page. It works.

harrislapiroff:
I.e., it's not overriding the version that django and grappelli use. It's giving them a version to use at all.

PR #331, opened 2015-02-09
Corrected event update form support for live/test apis. Resolved #329.

### LINKED ISSUE #329: Event update form doesn't fully support test events
There are two issues:

1. The test api fields are not excluded from the form, which means they get blanked every time the form saves.
2. If the "disconnect" boxes *are* checked, it always manually blanks the non-test fields. Which means that they'll get erased when they shouldn't, and that if the test fields *weren't* getting blanked, they never could be.

While we're fixing this, we should also implement *actual* oauth disconnects, rather than just deleting our copy of their data.

Original report was in #320.

melinath:
From the ticket:

There are two issues:

1. The test api fields are not excluded from the form, which means they get blanked every time the form saves.
2. If the "disconnect" boxes are checked, it always manually blanks the non-test fields. Which means that they'll get erased when they shouldn't, and that if the test fields weren't getting blanked, they never could be.

This pull request resolves both issues, resulting in a form that functions as it should.

### DIFF
--- a/brambling/forms/organizer.py
+++ b/brambling/forms/organizer.py
@@ -44,10 +44,13 @@ class EventForm(forms.ModelForm):
         model = Event
         # TODO: When multiple countries are supported, remove 'country'
         # and 'check_country' from excludes:
-        exclude = ('dates', 'housing_dates', 'owner',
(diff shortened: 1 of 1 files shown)


melinath:
@harrislapiroff Request for review

PR #339, opened 2015-02-11
Switch to django-talkback.

harrislapiroff:


harrislapiroff:
Alright, @melinath, this should be all good now.

PR #350, opened 2015-02-15
Remaining display

melinath:
Lets organizers choose on a per-option basis whether and how they want to display the number available. Even has a fancy collapse in the itemoption editing form!

### DIFF
--- a/brambling/migrations/0018_itemoption_remaining_display.py
+++ b/brambling/migrations/0018_itemoption_remaining_display.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('brambling', '0017_auto_20150214_0028'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='itemoption',
+            name='remaining_display',
+            field=models.CharField(default=b'both', max_length=9, choices=[(b'both', 'Remaining / Total'), (b'total', 'Total only'), (b'remaining', 'Remaining only'), (b'hidden', "Don't display")]),
+            preserve_default=True,
+        ),
+    ]
--- a/brambling/models.py
+++ b/brambling/models.py
@@ -398,12 +398,23 @@ class ItemImage(models.Model):
 
 
 class ItemOption(models.Model):
+    TOTAL_AND_REMAINING = 'both'
+    TOTAL = 'total'
+    REMAINING = 'remaining'
+    HIDDEN = 'hidden'
+    REMAINING_DISPLAY_CHOICES = (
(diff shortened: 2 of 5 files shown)


melinath:
This still needs a little work... dynamically-created itemoption forms don't work properly yet. Not sure of the best way to handle that.

nmorduch:
http://api.jquery.com/on/#direct-and-delegated-events is probably your problem

melinath:
@nmorduch The problem is more that the events are attached via id rather than via class - so I'd need a way to figure out what the new id would be. Better alternative is probably to use a class with data attributes... I guess?

melinath:
@nmorduch I was hoping you might be able to fix it.

nmorduch:
@melinath i'm a little lost in what already exists

and more importantly, it was time for dinner

nmorduch:
Also, actually, I would propose the addition of a "default" option, and then you can set the default somewhere else. Because although the granular control can be helpful, with many item having many options, it's going to be a pain to adjust this on every item.

melinath:
@nmorduch good idea. Do you think it should be event-level? Item-level? Both?

melinath:
We could put a dropdown on the item list page that lets you a) set the default, b) optionally set for *all* itemoptions. It could store on the event model but not show up on the event settings page...

OTOH that would be kinda close to an actions dropdown, which might be a little weird.

harrislapiroff:
How necessary is it to have the multiple different display options? If it was only show vs hide, here's an interface that I think would make sense:

- checkbox on the `Item` "Hide remaining counts"
- if the checkbox is not checked, show a checkbox on each `ItemOption` for "Hide remaining counts" (if the checkbox on the `Item` is checked, these boxes do not display at all)

I'm -1 on event level setting because I think it's too distant from itemoptions and people won't understand what it does. I'm also -1 on an action dropdown because I think it will be confusing and not used frequently enough to justify it.

nmorduch:
I like @harrislapiroff's option :)
Will that work for the event that requested this? Did they want to hide total or remaining?

melinath:
It's not just show vs. hide. There are at least three options:

1. Hide all
2. Show how many are remaining, but not the total
3. Show how many are remaining, out of the total

These three we should definitely support. The fourth case (show total, but not how many are remaining) could probably be eliminated IMO; I don't see a strong use case for it.

harrislapiroff:
Yeah, I mean, I saw those options in the code. I was just wondering if we could narrow it down to make configuration simpler.

melinath:
The request was specifically for "Show how many are remaining, but not the total", which I think is definitely a reasonable thing to want.

I guess we could eliminate the "remaining / total" option. Is that what you were proposing?

harrislapiroff:
Oh, hm. I see. Yeah, I'd tentatively propose that, but I can also understand why we might not want to get rid of that option. Hrf.

melinath:
Okay, based on our 2/16 discussion, merging this.

PR #374, opened 2015-02-28
Ticket/340

### LINKED ISSUE #340: External website button on shop pages
- [x] Should be called something more explicit (better than implicit) than "More information"
- [x] Should have `target="_blank"`

nmorduch:
Closes #340, which I forgot to say. Would be nice for the near future if non-dance-event doesn't put a description.


nmorduch:
ooops you did some of this already, @melinath. i'll look at yours.

PR #376, opened 2015-02-28
Organizer filter views

melinath:
Improvements to organizer filter views. Corrected CSV response generation to *actually* use the same table and implemented search.


harrislapiroff:
I'm a little concerned that putting the loupe button on the far side of the change columns and filter buttons is confusing. Is it okay if I move it next to the search bar?

harrislapiroff:
I'm also a little concerned that this form might have some unexpected behavior. For instance, if I click "Change columns and filters", make some changes, press "Cancel", then enter a search term and press the loupe, the changes I supposedly cancelled will take effect.

harrislapiroff:
Maybe I should just merge this as is and do a UI cleanup on another branch...

melinath:
I'm okay with you moving the loupe. But I think I do prefer the UI cleanup as a separate branch, potentially.

harrislapiroff:
:+1:

harrislapiroff:
Also, if I enter search terms or change the filters, don't submit, and then press download CSV, it'll download a filtered CSV. There's a lot of potential UI failure points in making this a single form. I'll think about how to best break it up for another ticket.

harrislapiroff:
@melinath Bouncing back to you for a final check. Merge if it's ready.

melinath:
Looks fine to me.

PR #377, opened 2015-03-01
Implemented admin-facing option to manually send order receipts.

### LINKED ISSUE #356: Add "send receipt" button to order detail page in admin
Occasionally, for whatever reason, a person may not get a receipt (or may lose it.) It would be good to have a handy button that organizers could just click to re-send that email.

melinath:
Resolved #356.

### DIFF
--- a/brambling/templates/brambling/event/organizer/order_detail.html
+++ b/brambling/templates/brambling/event/organizer/order_detail.html
@@ -40,7 +40,14 @@
 {% block main %}
 	{{ block.super }}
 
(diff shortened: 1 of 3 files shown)


harrislapiroff:
Merging this as-is, but wanted to float a thought about making action-based requests like this POST requests instead of GET. It doesn't seem like a _huge_ difference, but it feels technically correct to me, since GET responses are, in principle, supposed to be requests that return a cacheable resource and have no other effect. It might also give some extra security in terms of checking CSRF.

harrislapiroff:
(We could just make it a form with a hidden field and a button.)

PR #378, opened 2015-03-01
Improved invite acceptance process.

### LINKED ISSUE #373: Accept Invitation UI
The screen for getting invited to manage an event looks like this:

screenshot 2015-02-27 18 42 55

This can be confusing as a new user, especially since it defaults to Log In rather than Sign Up.

This UI should be improved. Here are some ideas:

* I'm not sure if invitations are (or should be) transferable (i.e., if you can sign up with a different email address than the one that received the invitation). If they are not, we should simply present the appropriate log in or (exclusive) sign up form appropriate to the email address.
* Otherwise, we could default to sign up, since experienced users are more likely to realize they need to switch to the log in screen.
* Either in combination with the above or instead of it, we could add some explicative text, e.g., "If you already have a Dancerfly account blah blah blah etc."

melinath:
Resolved #373.
Also slightly improved the email html and made Invite code swear-safe.

### DIFF
--- a/brambling/templates/brambling/invite.html
+++ b/brambling/templates/brambling/invite.html
@@ -8,20 +8,39 @@
 	If the person is already logged in, the invitation is currently automatically accepted.
 	{% endcomment %}
 	{% if invite %}
+		<h2>
+			You've been invited
+			{% if invite.kind == invite.HOME %}
+				to edit a home!
+			{% else %}
+				to edit an event!
+			{% endif %}
+		</h2>
+		<p>
+			This invitation was sent by <strong>{{ invite.user.get_full_name }}</strong> to <strong>{{ invite.email }}</strong>.
+			{% if invited_person_exists %}
+				Log in to accept the invitation!
+			{% else %}
+				Create an account to accept the invitation!
+			{% endif %}
+		</p>
+		{% if request.user.is_authenticated %}
+			<div class='alert alert-warning'>
+				<p>You are currently logged in as <strong>{{ request.user.email }}</strong>. You can only accept this invitation as <strong>{{ invite.email }}</strong>. If you believe this invitation was sent to the wrong email address, please get in touch with <strong>{{ invite.user.get_full_name }}</strong> directly to request a new invitation.</p>
+			</div>
+		{% endif %}
 		{% if invited_person_exists %}
-			<h2>Log in to accept the invitation from {{ invite.user.get_full_name }}</h2>
 			<form id="login" action="{% url 'login' %}?next={{ request.path }}" method="post">
 				{% csrf_token %}
 				{% form login_form %}
-				<button class="btn btn-primary-light" type="submit">Log in</button>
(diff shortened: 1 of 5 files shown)


melinath:
This should also resolve the error emails we got around invitations.

brambling/templates/brambling/invite.html
```diff
@@ -4,25 +4,26 @@
...
 				<button class="btn btn-primary-light" type="submit">Log in</button>
 				<a class='btn btn-link' href="{% url 'password_reset' %}">Forgot your password?</a>
 			</form>
-			<form id="signup" class="tab-pane fade" action="{% url "brambling_signup" %}?next={{ request.path }}" method="post">
+		{% else %}
+			<h2>Create an account to accept the invitation from {{ invite.user.get_full_name }}</h2>
+			<form id="signup" action="{% url "brambling_signup" %}?next={{ request.path }}" method="post">
 				{% csrf_token %}
 				{% form signup_form using "brambling/forms/signup.html" %}
 				<button class="btn btn-primary-light">Sign Up</button>
```
nmorduch: It is possible that the person will have an account but not have been invited using the same email address. Can we add a link to log in? Maybe after the signup button, a link saying "Already have an account? Log in."

melinath: Invites can only be accepted by an account that uses the email address that the invite was sent to.

melinath: We could change that, hypothetically, but I don't think we should.

nmorduch: Oh interesting. I don't think places typically do it that way? But that is easier.

nmorduch: What happens if they're currently logged in with a different email address?

melinath: They are presented with the same form as if they weren't logged in. *shrug*

nmorduch: Does that work? Can you log in to another account without logging out first?

melinath: Yes.

nmorduch: Cool.

melinath: It *would* be nice if we could prevent people from logging in / signing up with other email addresses, since it wouldn't work (and then they'd be confused) but I think this should cover most cases as-is.

nmorduch: Yeah this is going to make a case where:
- You're logged in
- You get an invitation to a different email address
- You get a log in screen so you log in with your existing log in
- Nothing has happened.

I think we should either be explicit about them being tied to email addresses (probably in the email itself) or not tie them to email addresses.

harrislapiroff:
We could also disable the email field and add help text clarifying that you're not allowed to claim the invitation as someone else.

harrislapiroff:
But that can be a separate ticket.

harrislapiroff:
Out of curiousity, though, what makes the invite code swear-safe? It looks like you made it safe for writing down unambiguously (which seems unnecessary, but fine)? Is that related?

harrislapiroff:
This looks good, though, so I'm gonna merge.

PR #392, opened 2015-03-05
Corrected removal of additional items from Attendees.

### LINKED ISSUE #387: Item Assignment Buggy
Non-pass item assignment seems to be sort of buggy. Naomi and I ran into a few things including:

* Having a single item show up for multiple attendees.
* Having an item show up with the message "all items in order must be assigned to an attendee" but no option to assign the item to an attendee.
* Just generally unpredictably and wonky behavior.

This whole view should probably be audited.

melinath:
Resolved #387.


harrislapiroff:
Seems to work! Mergin.

Issue #338, opened 2015-02-11
Fix z-index bugs during modals.


melinath:
There are some strange situations when modals pop up and some content is on top of the modal that shouldn't be. Simple example, when you're trying to refund an order.

![screen shot 2015-02-10 at 6 27 24 pm](https://cloud.githubusercontent.com/assets/299979/6140879/84c84d9e-b152-11e4-8e9a-3c2f27ee6f30.png)

Also, when you're shopping and run out of time.

![screen shot 2015-02-10 at 6 29 45 pm](https://cloud.githubusercontent.com/assets/299979/6140905/d7fb0d62-b152-11e4-8d2c-debc8092d0c9.png)

melinath:
Also, the close button in the modal title area (on order/attendee filter views) is *below* the modal title, which means it is unclickable.

nmorduch:
I made the second comment into its own issue #375 ^

nmorduch:
Note:

```scss
//-- Z-index master list
//
// Warning: Avoid customizing these values. They're used for a bird's eye view
// of components dependent on the z-axis and are designed to all work together.
//
// Note: These variables are not generated into the Customizer.
$zindex-navbar: 1000 !default;
$zindex-dropdown: 1000 !default;
$zindex-popover: 1060 !default;
$zindex-tooltip: 1070 !default;
$zindex-navbar-fixed: 1030 !default;
$zindex-modal: 1040 !default;
```

melinath:
So... did we customize them? What happened here?

nmorduch:
I might be better able to look into this if the "refund order" button worked for me :p

nmorduch:
The :recycle: button (i know it doesn't look exactly like that) did bring up a modal but with no backdrop

PR #399, opened 2015-03-06
closed #375 modal close button

### LINKED ISSUE #375: Filter modal close button
@melinath says: the close button in the modal title area (on order/attendee filter views) is below the modal title, which means it is unclickable.

From #338

nmorduch:
I'm not sure why this was a problem, since we use the example bootstrap markup, but this does fix it.

### DIFF
--- a/brambling/sass/modules/_modals.sass
+++ b/brambling/sass/modules/_modals.sass
@@ -1,0 +1,2 @@
+.modal-title
+	margin-right: $line-height-computed
--- a/brambling/sass/styles.sass
+++ b/brambling/sass/styles.sass
@@ -69,6 +69,7 @@
 @import modules/layout
 @import modules/masthead
 @import modules/menus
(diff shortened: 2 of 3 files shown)


melinath:
Isn't this just a z-index issue like #338? Slash, if so, is it a good idea to use a workaround rather than finding the root?

nmorduch:
I didn't see any z-index on it, but yes you're right that it would be better to find the root. But I also looked and did not find it.

harrislapiroff:
The difference between Bootstrap's default and ours is that all of our titles (h1...h6) are positioned relatively (in order to fix the st ryde baseline). This implicitly changes the stacking order because relative elements are (I think) always stacked on top of static ones. This seems like a good enough solution. An alternative option would be to position the close button relatively and fuck with its z-index, but that doesn't seem cleaner.

PR #401, opened 2015-03-06
Corrected workflow for SurveyDataView.

### LINKED ISSUE #400: Survey view can't be accessed post-checkout
Clicking on the "survey" button in the workflow listing just redirects you back to the summary view. This is not good. :-p

melinath:
Resolved #400.


harrislapiroff:
This branch seems fine, but in my testing I've noticed a lot of buggy/unexpected behavior in the step bar, not all of which I know how to reproduce consistently. I'll try to create some tickets for stuff.

PR #404, opened 2015-03-10
Ticket/243 (Custom forms, stage 1)

melinath:
This implements custom forms for attendees and orders. Supported field types: text, boolean, textarea. Data is stored as JSON in the db, which means we won't easily be able to filter on it (not that we would be able to do that easily anyway...)


brambling/models.py
```diff
@@ -1361,3 +1370,129 @@ def send(self, site, body_template_name='brambling/mail/invite_{kind}_body.html'
...
+        (ATTENDEE, _('Attendee')),
+        (ORDER, _('Order')),
+
+    )
+    form_type = models.CharField(max_length=8, choices=FORM_TYPE_CHOICES,
+                                 help_text='Order forms will only display if "collect survey data" is checked in your event settings')
+    event = models.ForeignKey(Event, related_name="forms")
+    # TODO: Add fk/m2m to BoughtItem to limit people the form is
+    # displayed to.
+    name = models.CharField(max_length=50)
```
harrislapiroff: Help text? "This name is for organization purposes only and will not be displayed to attendees."

brambling/forms/organizer.py
```diff
@@ -265,6 +266,30 @@ def _post_clean(self):
             self._update_errors(e)
 
 
+class CustomFormForm(forms.ModelForm):
+    class Meta:
+        model = CustomForm
+        exclude = ('event',)
```
harrislapiroff: also `'index'`?

brambling/templates/brambling/event/organizer/custom_form_form.html
```diff
@@ -0,0 +1,110 @@
...
+				deleteTrigger: '<a href="#"><i class="fa fa-times"></i></a>',
+				deleteWrapper: '.actions',
+				addTrigger: '<a href="#"><i class="fa fa-fw fa-plus"></i> Add a field</a>',
+				addWrapper: '.js-formset-add.option-add',
+				formTemplate: '.field-template',
+				sortableHandle: '.panel-heading',
+				sortableField: 'index'
+			});
+		}());
+	</script>
```
harrislapiroff: Some wonky stuff is going on in this formset. In particular, dropdowns aren't working with dynamically created items.

brambling/templates/brambling/event/organizer/custom_form_list.html
```diff
@@ -0,0 +1,33 @@
...
+	{% include "brambling/event/_empty_shop_alert.html" %}
+
+	<p><a class="btn btn-primary" href="{% url 'brambling_form_create' event_slug=event.slug %}">Create form</a></p>
+	{% if custom_forms %}
+		<table class='table table-striped'>
+			<thead>
+				<tr>
+					<th>Name</th>
+					<th>Type</th>
+					<th>Index</th>
```
harrislapiroff: I don't think users are going to understand what this field is. _I_ don't understand what this field is.

harrislapiroff:
After an hour of testing I only discovered the one javascript bug listed above and the various UI thoughts I commented on. Good work! Surprisingly (so far) bug free for a major new feature!

:hocho: :bug:

harrislapiroff:
I'm gonna look over the other PRs and then come back to this and see if I can fix the bugs and UI stuff.

brambling/templates/brambling/event/organizer/custom_form_form.html
harrislapiroff: Although, actually, the ideal interface would be more like other form builders and not let the user think about a field type as mutable. I.e., they would create a text field and if they wanted to change it to a checkbox, they would delete it and create a checkbox instead.

That should probably be for another PR though.

brambling/forms/organizer.py
melinath: As in... you think we should hide index for now?

brambling/models.py
melinath: If people have multiple forms for Order and Attendee, we might want to display the name someday.

melinath: At least, that's what I was thinking.

harrislapiroff: Well, we could still add the help text to describe the current behavior. If we change it for multiple forms, we probably still won't change it for single forms, so people won't be misled.

brambling/forms/organizer.py
harrislapiroff: I think so? Why would I ever change that as an event organizer under the current behavior?

brambling/templates/brambling/event/organizer/custom_form_list.html
melinath: Index? The forms are displayed in an order... i.e. if there are multiple Order forms, they would be displayed in index order. Someday that will translate to which page they are on.

brambling/forms/organizer.py
harrislapiroff: If there's no current use for them, it's going to confuse users. We should definitely exclude it IMO.

harrislapiroff: I say this because it confused me and I had the advantage of having already looked over the code.

brambling/templates/brambling/event/organizer/custom_form_list.html
harrislapiroff: ditto from the comments above on excluding the index field ^

brambling/models.py
melinath: (We would also display it if there were multiple pages of forms.)

melinath: But yeah.

brambling/forms/organizer.py
melinath: Okay.

brambling/models.py
```diff
@@ -1361,3 +1370,129 @@ def send(self, site, body_template_name='brambling/mail/invite_{kind}_body.html'
...
+        return field_class(**kwargs)
+
+
+class CustomFormEntry(models.Model):
+    related_ct = models.ForeignKey(ContentType)
+    related_id = models.IntegerField()
+    related_obj = GenericForeignKey('related_ct', 'related_id')
+
+    form_field = models.ForeignKey(CustomFormField)
+    value = models.TextField(blank=True)
```
harrislapiroff: Thought for the future: Could we use PostgreSQL's hstore or json type for this?

melinath: yeah, probably? I don't know much about them / integrating them with Django.

brambling/forms/organizer.py
harrislapiroff: Oh, I see. We already do support multiple forms. Hrm. :|

harrislapiroff: I think the UI for multiple forms needs to be pretty heavily rethought and maybe even the data structures. It doesn't make sense to have both attendee and event forms in the same table if you're going to be reordering them (or even thinking about their order) in that view.

harrislapiroff: Maybe I'll just do a groupby in the template for now.

harrislapiroff: This particular line will stay the same, though, for now. This issue is resolved by the addition of help text.

brambling/models.py
harrislapiroff: Resolved.

brambling/templates/brambling/event/organizer/custom_form_form.html
harrislapiroff: This bug actually seems to be on master.

harrislapiroff:
Okay, I think I'm done with everything I want to do. Bouncing back to you @melinath.

brambling/templates/brambling/event/organizer/custom_form_form.html
melinath: +1 on not having the field type mutable.

melinath:
Looks good.

Issue #209, opened 2014-10-03
Clean up behavior of "next step button"


nmorduch:
Right now, it sometimes looks like it should have submit button functionality when it does not. Also, it and the countdown should maybe not be at the top of the page. See https://github.com/littleweaver/django-brambling/commit/37f4f8a2263e#commitcomment-8032523 and nearby comments

harrislapiroff:
My going theory is to make more of the views single column and have a static side column that has the timer, and the next step button.

harrislapiroff:
Gonna try it.

harrislapiroff:
Resolved by bf65fd06185e83ff208f3b4bbfef08b0beb1756b

PR #408, opened 2015-03-13
Next step button on attendee basic information page now submits the form.

harrislapiroff:
Instead of going to the next step, pressing "next step" submits the form. I believe this should be harmless, even if the form is unchanged, and will result in more predictable behavior. Partially resolves #209.

PR #414, opened 2015-03-18
Checkout relayout

harrislapiroff:
I updated the layout for all the order pages. In particular:

* The next step button never appears at the top of a page
* A cart button appears up there instead to remind you how many items you are buying
* The cart button prompts a modal that displays all the items in your cart
* The next step button _almost_ always will appear on the bottom right of the page, except when editing an attendee where a user must explicitly choose to save the attendee or cancel
* The next step button is always at the bottom on mobile!
* This whole thing works much better on mobile, I think, though the stepbar still needs love.

There's more that could be done in terms of polish here, but I think it's also ready to go as-is. I'm open to suggestions for this PR or to future polishing tickets.

* [x] Remove javascript in modal doesn't work.
* [x] Hosting page sucks.
* [x] Housing page sucks.

### DIFF
--- a/brambling/templates/brambling/event/order/__base.html
+++ b/brambling/templates/brambling/event/order/__base.html
@@ -114,5 +114,33 @@
 	{% endif %}
 
 	{{ block.super }}
+
 	<script type="text/javascript" src="{% static "zenaida/js/bootstrap/modal.js" %}"></script>
+
+	{# Make the shopping cart remove button ajaxy. #}
+	<script type="text/javascript">
+		{# TODO: Rewrite this javascript not to use the `target` attribute. Probably to submit forms. #}
+		$(function(){
+			$('a[target]').on('click', function(e){
+				var $this = $(this);
+				if ($this.attr('target') === '_blank') return true;
+				$.post($this.attr('target'), function(data){
+					if (data.success) {
+						// if it's a remove button, don't reload, just remove the item
+						if ($this.is('[target*="remove"]')){
+							$this.closest('li, tr').remove();
+						// otherwise, just refresh for now
+						} else {
+							document.location.reload()
+						}
+					} else if (data.error) {
+						$('.stepbar').after(
+							"<div class='alert alert-dismissable alert-danger'><button type='button' class='close' data-dismiss='alert' aria-hidden='true'>&times;</button> " + data.error + "</div>"
+						);
+					}
+				});
+				e.preventDefault();
+			})
+		});
+	</script>
 {% endblock %}
--- a/brambling/templates/brambling/event/order/__base.html
+++ b/brambling/templates/brambling/event/order/__base.html
@@ -92,9 +92,11 @@
 
 {% block after %}
 	<div class="row">
-		<div class="col-sm-4 col-sm-offset-8">
+		<div class="col-md-5 col-md-offset-7">
 			{% block next_step_button %}
-				{% include "brambling/event/order/_next_step_button.html" %}
+				{% if next_step %}
+					{% include "brambling/event/order/_next_step_button.html" %}
+				{% endif %}
 			{% endblock %}
 		</div>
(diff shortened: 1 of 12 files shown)


harrislapiroff:
This supersedes #408 if merged.

harrislapiroff:
screenshot 2015-03-18 13 55 48

harrislapiroff:
This also resolves #209

brambling/templates/brambling/event/order/__base.html
```diff
@@ -47,8 +47,39 @@
...
+													<a href="#" class="pull-right" data-dismiss="modal" aria-label="Close"><i class="fa fa-times"></i></a>
+													<h4 class="modal-title" id="cartTitle">Cart ({{ cart.count }} item{{ cart|pluralize }})</h4>
+												</div>
+												<table class="table">
+													{% for options in cart_list %}
+														{% for personitem in options.list %}
+															<tr>
+																<td>{{ personitem.item_option.item.name }} ({{ personitem.item_option.name }})</td>
+																<td>{{ personitem.item_option.price|format_money:event.currency }}</td>
+																<td><a target='{% if code_in_url %}{% url "brambling_event_shop_remove" event_slug=event.slug pk=personitem.pk code=order.code %}{% else %}{% url "brambling_event_shop_remove" event_slug=event.slug pk=personitem.pk %}{% endif %}' href='javascript://' class='text-danger pull-right'><span class='fa fa-times'></span></a></td>
```
nmorduch: JS on this isn't working for me. Opens a new tab.

nmorduch:
Not only do we still have the stepbar bug, but this is butt-ugly

nmorduch:
As is this 


And I'd like to change this before merging because I don't want to be changing the UI every 10 seconds.

nmorduch:
It's good other than the alignment of the button, though. And that includes the cart button compared to the timer.

harrislapiroff:
"And I'd like to change this before merging because I don't want to be changing the UI every 10 seconds."

Your specific complaints seems fine, for completing this branch, but does this mean that you prefer large design overhauls to incremental UI improvements in general? I'd personally rather make frequent small changes to the UI, but we could talk it out.

harrislapiroff:
Also, wrt the specific examples you gave, do you have specific solutions in mind?

nmorduch:
I knew that was unclear when I said it. Should have clarified then.
I don't want to move buttons around all the time. If you're a user asking the someone for help and they say "and then look for the button in the upper left" I don't want it to have moved three times since they last looked at it. That's my only specific worry about incremental changes. Building on things is good. I just don't want to move things around too often or switch them on and off too often.

re specific examples:
- Cart button height is non-urgent but we should have a button class that's specifically for going next to an alert.
- Next step button should be in the left column with the form, or should be full-width, or left-aligned, or something. Not sure.

harrislapiroff:
@nmorduch Cart button height issue is also on master (with the other button instead) so I might prefer not to deal with it on this branch. If the next step button were left aligned, would you want that to be on all pages or just the ones that have narrow forms?

nmorduch:
1. fair
2. I don't know, probably on all? consistency is good. But what do you think?

harrislapiroff:
@nmorduch I've fixed the three issues listed above.

The fix to the cart is not ideal, but a fuller fix will require much more time, so I think it should be a separate ticket. Unfortunately implementing cart editing on pages that are not the shop is prone to lots of weird behaviors (what happens if you remove all the passes? what happens if you remove the pass that you are currently editing? what if you're on the housing screen and you remove the only pass that needs housing? etc.) that require a more robust and less ad-hoc ajax framework to handle. Instead I've simply removed the ability to edit items in the cart and added a note directing users back to the shop page if they wish to do so.

nmorduch:
v. good. did you make a ticket for cart editing?

PR #432, opened 2015-03-25
Ticket/409

melinath:
This branch makes the ModelTable a more standalone class. It handles all filtering, ordering, and searching internally, instead of having some of that handled in the view and some in the ModelTable / FilterSet.

It also moves the CSV generation into the view, since that is more presentation-related (as opposed to data manipulation.)

### DIFF
--- a/brambling/templates/brambling/event/organizer/__table.html
+++ b/brambling/templates/brambling/event/organizer/__table.html
@@ -56,7 +56,7 @@
             <thead>
                 <tr>
                     {% for header in table.header_row %}
-                        <th>{{ header }}</th>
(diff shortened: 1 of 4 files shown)


harrislapiroff:
Something's not going right with field names. Some of the housing fields are just labelled "Inner":

master | ticket/409
---------------|---------------
screenshot 2015-03-31 12 32 45 | screenshot 2015-03-31 12 33 02

harrislapiroff:
Everything else seems to otherwise be working. I don't understand enough about the code to fix this bug myself, so flipping back to you @melinath.

harrislapiroff:
Looks good.

Issue #409, opened 2015-03-15
Refactor ModelTable to more closely match ModelAdmin / incorporate filters


melinath:
The ModelTable interface is kinda a mess. For one thing, config is all over the place - filtering happens via FilterSet, searching happens in the view, and what fields are available happens on the table. Which wouldn't necessarily be as weird except that the table also affects the queryset - which means that we could (hypothetically) need to set up a FilterSet to filter on a value created by the ModelTable. While this is certainly possible, and probably fine for now, I would like to propose the following changes:

* *ModelTable.list_display:* A full list of all of the fields that can be included in the table. Potentially plus a `get_list_display` method.
* *ModelTable.default_selected:* A full list of all fields that are included by default. This could be a property and default to the full list, or could be paired with a `get_default_selected` method.
* *ModelTable.field_groups / fieldsets:* A dictionary / tuple-tuple of field groupings which (someday) can be activated / deactivated at once (#219), and which (for now) can be treated as display groupings for the form.
* *ModelTable.search_fields:* A tuple of fields / relations that will be searched for the given value.

I think this will be cleaner to read, and will also be easier for other folks to use if we ever split this off into a separate app. However, it does lose the ability to easily define new table header text for the various fields.

melinath:
Additional information:

ModelTable currently calculates its final QuerySet in `__iter__`, which means (among other things) that a FilterSet defined in the view **can't** filter on any annotated values set up in `ModelTable.get_queryset()` (because they don't exist yet.) Part of me wants to make ModelTable completely agnostic as regards the queryset - i.e. make it just assume that the fields are all there and move all queryset handling into the view - but I think it would make more sense to move all the qs handling into the table itself.

melinath:
I think having a label_overrides dict would solve the problem of defining new table header text nicely.

melinath:
This was merged in as commit b24ac0ce6194f0118765f620bb25e0ad83c38f1d.

melinath:
PR #432

PR #436, opened 2015-03-27
Added 'notes' tab to order management for admins.

### LINKED ISSUE #242: Arbitrary notes field on orders and attendees
Organizers should be able to edit an arbitrary internal text field of notes on each order and attendee. This will make the brambling database more extensible as it will allow them to annotate each record with whatever data they need to keep track of.

This data should also be searchable, probably.

melinath:
Related to #242. Really, I think this should close #242. Adding notes field to attendees requires a much more invasive change, since we don't even have a way to view them in the admin right now.

Part of me wants to prioritize this, since we have a request from an event for notes functionality - but they might be looking for notes on attendee, not on order. Not sure.

This also needs to be added to the model table display, but I didn't want to much up the merge of the #409 branch by creating conflicts... Maybe I shouldn't worry about that, though.

### DIFF
--- a/brambling/forms/organizer.py
+++ b/brambling/forms/organizer.py
@@ -8,7 +8,7 @@ import floppyforms.__future__ as forms
 
 from brambling.models import (Attendee, Event, Item, ItemOption, Discount,
                               Date, ItemImage, Transaction, Invite, CustomForm,
-                              CustomFormField)
+                              CustomFormField, Order)
 from brambling.utils.international import clean_postal_code
 
 from zenaida.forms import (GroupedModelMultipleChoiceField,
@@ -355,3 +355,9 @@ class ManualDiscountForm(forms.Form):
 
     def save(self):
         self.order.add_discount(self.cleaned_data['discount'], force=True)
+
+
+class OrderNotesForm(forms.ModelForm):
+    class Meta:
+        model = Order
+        fields = ('notes',)
--- a/brambling/migrations/0025_order_notes.py
+++ b/brambling/migrations/0025_order_notes.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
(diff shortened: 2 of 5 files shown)


nmorduch:
Looks cool to me but I'll flip it to Harry B for the python. I agree about putting it in the table, so let's close #242 and make a new issue for showing notes in the table

harrislapiroff:
I'm going to look over this and (probably) merge it anyway, but I have a lot of concerns about attaching increasing amounts of data (between this and the surveys) to Order objects.

I think an Attendee is the fundamental object of dance event management and an Order is an implementation detail that should recede into the background as much as possible. I have trouble thinking of too many circumstances in which attaching data to an order is more useful than attaching data to the attendee instead and I think we invite confusion about where data belongs by putting it in both places (or just on the Order, which I think of as the _wrong_ place for this sort of data).

nmorduch:
I agree about this. Problem is that minimum option is typically order?

harrislapiroff:
I mean, not if the order is not part of the feature at all.

harrislapiroff:
What I mean to say is that if I were designing Dancerfly on my own, for instance, all data would be associated with the attendee and the Order would be _exclusively_ a financial record that was associated with certain bought items. There would be no data beyond that associated with the order.

nmorduch:
Fair.

harrislapiroff:
This is also to say that I don't think this closes #242, because I think it's way less useful and covers way fewer use cases IMO than 242.

harrislapiroff:
This seems good though and bug free. Merging.

harrislapiroff:
(Also, acknowledging that I did apparently include order notes in the ticket, so this is my personal change of heart that's driving this conversation.)

melinath:
I am also now in a place where I would also probably have designed it somewhat differently if I were doing it now. We can move in that direction, slowly.

PR #440, opened 2015-03-31
Compressor

harrislapiroff:
This branch should be ready for testing, though I'd like to make sure we test it pretty heavily before merging, since it's a major change. It would be nice to do the following, but maybe is not required for this merge?

* [ ] find a more robust way of adding bootstrap-sass's `assets/stylesheets` file to add to `STATICFILES_DIRS` in `settings.py`
* [ ] find a way to include Autoprefixer
* [ ] find a way to install `bootstrap-sass` as part of the python installation process instead of having to use Bundler separately


melinath:
I actually think the inclusion is reasonably robust. We just have to keep the versions in sync. I seriously doubt we'll find a way to manage ruby gems with Python and I doubt it's worth the time. Can't speak to autoprefixer, but I imagine it would be pretty simple?

It looks like we aren't currently compressing js - any particular reason?

melinath:
Oh, it's because my debug is True.

melinath:
There are some files that aren't getting caught by the compression that could be - for example, bootstrap carousel.js and modal.js, and St Ryde stylesheets. Would we want to just have those always available, but compressed? Or is it not worth the action-at-a-distance effect?

melinath:
(This was just on a single page. Other one-off uses of various scripts probably also exist.)

melinath:
This doesn't seem to have a significant effect one way or the other locally, but I imagine the effect will be more noticeable remotely.

melinath:
I'm fine with merging. @nmorduch?

harrislapiroff:
One thing that definitely would be good to test (I tested it a little, but I want to be sure) is that during development, your changes to the SASS take effect immediately or in a timely manner.

harrislapiroff:
Also, I think we can rearrange where the compress tags are later on, but, yeah, it might make sense to move stuff that we use frequently inside the compress tags to the base template.

harrislapiroff:
@nmorduch Look at this pretty please?

test_project/test_project/settings.py
```diff
@@ -142,6 +145,20 @@
...
+    'django.contrib.staticfiles.finders.AppDirectoriesFinder',
+    # other finders..
+    'compressor.finders.CompressorFinder',
+)
+
+_BOOTSTRAP_SASS_SUBPATH = "/gems/bootstrap-sass-3.3.4.1/assets/stylesheets/"
+STATICFILES_DIRS = [x + _BOOTSTRAP_SASS_SUBPATH for x in os.environ['GEM_PATH'].split(":") if os.path.isdir(x + _BOOTSTRAP_SASS_SUBPATH)]
+
+COMPRESS_PRECOMPILERS = (
+    ('text/sass', 'django_libsass.SassCompiler'),
```
nmorduch: Why do we need this? The example in the docs has `('text/x-sass', 'sass {infile} {outfile}')`. (I'm wondering about sass compiler settings, because it would be nice to have it compile uncompressed in development.)

nmorduch:
I'm not finding that my Sass changes are taking quickly. But you say it's not caching when debug is True?

test_project/test_project/settings.py
harrislapiroff: I believe either mimetype is acceptable. Sass isn't responsible for compression. I agree that it'd be nice not to compress in development (I think Stephen turned it on). You just need to switch COMPRESS_ENABLED to False. Or maybe just set it to not DEBUG or something.

melinath: `COMPRESS_ENABLED = False` would be better than turning off DEBUG, since that has other side effects.

melinath: I had turned it on just to make sure that things actually worked with it on (before pushing). I think it's useful to be as close to production as possible, though I'd be fine with it off.

nmorduch: I was thinking of making it depend on whether debug is on. I think that's a thing you can do?

melinath: If you remove the setting altogether, it defaults to the opposite of DEBUG IIRC.

Issue #430, opened 2015-03-23
Compressor + Libsass + Autoprefixer


harrislapiroff:
It might be time to see if we can get completely away from using Compass and then successfully use Compressor with Libsass and Autoprefixer to automatically compile CSS on the fly.

Some potential leads:

* [Torchbox's Compressor Libsass processor](https://github.com/torchbox/django-libsass) - This looks reliable.
* [Django Compressor Autoprefixer](https://pypi.python.org/pypi/django-compressor-autoprefixer/0.0.1) - This doesn't...

harrislapiroff:
This functionality is currently active on #440.

melinath:
Merged in with ece735235ed6e99e7f98cba346f3a67dbacb52df.

Issue #385, opened 2015-03-02
Switch Order table to provide more annotated information


melinath:
Right now, we have a status field on Order which has a few options: IN_PROGRESS, PENDING, COMPLETED, and REFUNDED.

This is okay, but we can / should do better. For example, we now fully support multiple transactions on an Order. So an order can be "completed" and still have a cart. Or it can be "refunded", in that it has a refunded transaction, but "completed" in that it has another transaction that is still valid.

I think the solution is that we should instead annotate the queryset with the payment / refund information and use that to determine what state the order is in. Ideally in a way that lets us still filter by "state", but remove the status field.

melinath:
(this is waiting on the merge of #409 changes.)

PR #442, opened 2015-04-02
Added table count. Resolved #429.

### LINKED ISSUE #429: Add result count to model tables
It would be useful to have a result count (out of total, if filtered) display somewhere on the model table pages.

harrislapiroff:


melinath:
From a design perspective, it's a little weird that the baseline doesn't match up with the form elements. I'm guessing we're holding off on that?

From a functionality perspective, the link provided on the orders page to a full list of orders doesn't actually do so, because the default behavior for the orders page (if no parameters are provided) is to filter out in progress / refunded orders. This may be obviated somewhat once #385 lands - or we might end up having a similar setup. I think it would be nice to have a way to explicitly say what the "full listing" url would look like?

harrislapiroff:
Yeah, I don't wanna make too many changes to the sass before #430 is resolved and I think the best aesthetic approach would be to use a navbar (as noted in the comment) that was closer to bootstrap's default navbar.

Functionality-wise, you're right. Can you think of the best way to build this?

melinath:
I'll take a stab at it and then bounce this back to you.

melinath:
Unless you want to do it.

harrislapiroff:
FWIW what occurred to me was generating a version of the filterset form that has all hidden fields and blanks them all.

harrislapiroff:
Nah, you can do it however you think is good. I think you'll do a better job.

melinath:
Yeah, that was what I went with. Had to open gregmuellegger/django-floppyforms#145, but I was able to work around it in the meantime.

This change has introduced some new issues with UI, so it might be worth just fixing this as a whole... bouncing back to you.

harrislapiroff:
I made it palatable. I'll make a ticket to clean up the UI later. Merging.

Issue #382, opened 2015-03-01
Add support for dwolla transaction status changes.


melinath:
I believe this webhook handles transaction status changes that happen out-of-band. For example, if a payment is "reclaimed." Similar to #381 - charge disputes?

melinath:
This is getting rolled into #453 because this webhook gets used for ACH payments.

PR #455, opened 2015-04-15
Dwolla funding source support

melinath:
This pull request adds support for dwolla funding sources. Essentially, we check what funding sources are available when they pull up the payment form, then let them choose one to use. We also add support for the status change webhook (#382) because payments transferred via ACH will take a few days to clear.


harrislapiroff:
I tried to check this on an event with test APIs enabled and received a blank error upon checkout. I know the Dwolla test API does not function ideally or predictably (what even is the point of it tbh) but can you confirm that this is the expected behavior?

I used LW's sandbox account which I know does not have a funding source connected, which may also be part of the issue.

screenshot 2015-04-16 11 00 49

(Also as a sidenote, because the summary page is so long, I should probably put an error at the _top_ of the page for when payment fails. It's pretty confusing otherwise.)

melinath:
@harrislapiroff you probably need to update your dwolla version.

harrislapiroff:
I pip installed from requirements upon checking out this branch and the version I have installed matches the version in the requirements file (2.0.7).

melinath:
Hmm.

melinath:
@harrislapiroff Is the event also using LW's sandbox account? You can't send money from an account to the same account. That doesn't explain why the blank error, but it could explain why the error period?

harrislapiroff:
Yes, both are on the same account. That would make sense. Do you recommend I test this in some other way?

melinath:
Blank error is due to dwolla/dwolla-python#26.

PR #464, opened 2015-04-24
Date dates

### LINKED ISSUE #122: Refactor event start/end date handling / remove Date model.
Right now, events have an M2M to a Date object in the database (which just contains a date field). I don't think we actually need this for anything; pretty sure it's a case of overengineering on my part. We should remove it.

melinath:
Resolved #122. We should maybe even be going farther eventually – maybe storing an actual datetime of the event start (for more accurate sorting, for example) – but this is enough for now.


melinath:
(This pull request removes the dependencies between Event and Date / HousingSlot and Date. This eliminates the necessity for the .with_dates() annotation being sprinkled all over the place.)

PR #476, opened 2015-04-27
Marketing improvements

melinath:
This PR combines the splash & dashboard pages, transforms the About page into a more marketing-style page with calls to action and stuff, and adds default OG text to all pages that don't explicitly override it.


melinath:
Header for new dashboard page (dismissable for 1000 days, currently.)

screen shot 2015-04-27 at 12 06 47 pm

melinath:
About page screenshots.

screen shot 2015-04-27 at 2 14 51 pm
screen shot 2015-04-27 at 2 15 02 pm

harrislapiroff:
I like the direction of this a lot! I might even be ready to merge it and iterate from there. Some thoughts for now or later (I'll take a look at this actually on my computer soon):

* The element widths appear to be all over the place which is making me feel weird.
* The headlines are not very descriptive.
* Maybe we should advertise pricing on the same page and also give specific payment processor details. The text currently is pretty vague and we could probably spice it up with specifics.
* Speaking of specifics, let's advertise our specific features that set us apart from other ticketing services: Merch and Ticket sales in the same portal! Special features for housing coordination! Custom data collection! The features listed so far could be talking about almost any ticketing service.
* I donno about all the exclamation points.
* I donno about the semicolon either. (The usage is correct; I'm just not sure about semicolons in marketing copy.)

For a little more inspiration, here's Eventbrite's marketing page: https://www.eventbrite.com/how-it-works/ Some things that might be worth taking from them:

* Did you know that "Free for free events" was a feature we could advertise and not just a practical detail??? Hell maybe we could even mention it's free for events that take payment by mail!

brambling/templates/brambling/dashboard.html
```diff
@@ -4,43 +4,68 @@
...
+					<div class="col-sm-6">
+						<p class="text-jumbo">Start using Dancerfly for your <strong class='text-accent'>organization</strong> or <strong class='text-accent'>event</strong> today!</p>
+						<p>
+							<a class='btn btn-lg btn-primary-light'>Learn more <i class='fa fa-chevron-right'></i></a>
+							<a id='hideCall' class='btn btn-lg btn-link masthead-link'>Hide</a>
+						</p>
+					</div>
+				</div>{# /.row #}
+			</div>
+		{% endif %}
```
harrislapiroff: I think this reproduces what zenaida contrib hints is for.

brambling/templates/brambling/dashboard.html
```diff
@@ -92,24 +117,45 @@ <h4 class="panel-title">Past Events You Registered For</h4>
...
+			document.cookie = name+"="+value+expires+"; path=/";
+		}
+
+		$(function() {
+			$('#hideCall').on('click', function() {
+				createCookie('hideCall', 1, 1000);
+				$('.masthead').slideUp();
+			});
+		});
+	</script>
```
harrislapiroff: [v.s.](https://github.com/littleweaver/django-brambling/pull/476/files#diff-ca7a4ee4424afa2c841a2f979648168dR24)

brambling/templates/brambling/dashboard.html
harrislapiroff: Oh, hm. contrib hints requires a user. It should probably also accept a session.

harrislapiroff:
Also, I think maybe the full LW logo would be better than the square one here?

melinath:
Okay, I've made some changes - added payment section, switched to full logo, made the initial call to action clearer, a few other things.

harrislapiroff:
This looks much better! I still think you're emphasizing the wrong things on the marketing page though. "Easy checkout", "Easy data", and "Easy teams" are vague and could be said about almost _any_ ticketing solution out there and are unlikely to make anyone understand why they should use Dancerfly instead of e.g., Eventbrite, EventEspresso, et al.

It might be worth making a document similar to our LW content document, but I think more specific and unique things like "Get paid immediately" and "Housing coordination built-in" are going to be more compelling. If we can put one feature on that list that an organizer has been _hunting_ for, we can hook them.

Alternatively, if we can substantiate vague things like "Easy checkout", "Easy data", and "Easy teams" with screenshots or animations, I think that would improve them. But I'm not sure our UI is so beautiful yet that it sells itself.

That having been said, this is otherwise great and I'll merge is and submit content ideas as a separate PR.

PR #478, opened 2015-04-28
Order status purge

### LINKED ISSUE #477: Remove Order.status – Followup for #445
Order.status is a vestigial field now, which we don't really want to use. We need to go through and rip out the remaining references.

melinath:
Resolved #477. We mostly removed the need for Order.status a while ago when we switched to annotations for the table views, but we were still using it in a few places, which cluttered the code base and was muddying the waters for me re: pass transfer. So I shaved the yak.

This does contain a slightly unrelated change of factoring out workflow-specific functionality from the OrderMixin, but I think that will help in the long run, esp. with separating the checkout process from the summary view / potentially factoring workflow functionality off into a more self-contained utility.


harrislapiroff:
`/brambling/utils/model_tables.py:548` needs to be rewritten without using `event.housing_dates` but I'm not sure offhand how to do that. @melinath?

melinath:
Hmm, that should've been a part of date-dates, though, right?

harrislapiroff:
Thought for a future ticket:

We should clarify on the attendee table what the different possible statuses are and what they mean.

harrislapiroff:
This seems to work and I haven't run into any bugs.

PR #482, opened 2015-04-29
Org static slug

melinath:
This commit is basically a cleanup of things related to demo organizations. It provides two simple methods – `Organization.is_demo()` and `Event.is_demo()` – which can be used to check whether the organization or event is a demo.

In practical terms, this PR results in a few extra bits of helpful text being displayed, slugs not being modifiable, and new events being set to use test APIs.

It *might* make more sense to handle the demo slug as a setting rather than an attribute on Organization, but this is easier and it's not really ever going to be customized...


harrislapiroff:
Looks good.

Issue #460, opened 2015-04-22
Stop breaking up event list.


harrislapiroff:
Breaking up the event list on the dashboard and org pages (especially on org pages) looks weird and seems unnecessary. I'd prefer to have a combined list that makes certain to include all upcoming events that a user is registered for as well as a few that the user is not registered for and indicates visibly (I'm picturing a green button that says "Registered!" and replaces the current "Edit registration" button) which events you are registered for.

melinath:
Wouldn't that make it harder to find just events you registered for? I guess it doesn't matter *so* much since they'll hit the top of that list once they're most relevant...

harrislapiroff:
At least on org pages this is a necessity because the current layout looks silly and is confusing with so few events. (Also I'm not sure if this is confusing or broken, but under "Events You Registered For" _all_ events I'm registered for show up, not just the ones for that organization.)

I think I also prefer a combined feed for the homepage, though I'm open to persuasion. I do want to spend a little time thinking about how to call out an access the events you are registered for. Maybe they can also be listed in a separate location.

melinath:
+1 on changing it on the org pages, for sure. (That's broken.)

I like the idea of just having a separate page for events that you're registered for (or maybe only your events on the dashboard, and a separate page for browsing events).

Alternate options: we could look into implementing a full search page, and one of the options could be "events I'm registered for". But if we go in that direction, we'd need to somehow annotate the fact that someone is registered for an event, which could end up being a pain...

harrislapiroff:
I might also be into a combined list that had events you were registered for on top—basically what we have now but without headings or repetitions.

melinath:
This was resolved with the merge of #485.

PR #485, opened 2015-04-29
Combined upcoming event feed

harrislapiroff:
Events are now visually in a single feed and the update registration button indicates your registration state:

screenshot 2015-04-29 14 54 18

See #460

### DIFF
--- a/brambling/views/core.py
+++ b/brambling/views/core.py
@@ -43,10 +43,11 @@ class DashboardView(TemplateView):
 
             # Registered events is upcoming things you are / might be going to.
             # So you've paid for something or you're going to.
-            registered_events = list(Event.objects.filter(
+            registered_events_qs = Event.objects.filter(
                 order__person=user,
                 order__bought_items__status__in=(BoughtItem.BOUGHT, BoughtItem.RESERVED),
-            ).with_dates().filter(start_date__gte=today).order_by('start_date'))
+            ).with_dates().filter(start_date__gte=today).order_by('start_date')
+            registered_events = list(registered_events_qs)
             re_dict = dict((e.pk, e) for e in registered_events)
             orders = Order.objects.filter(
                 person=user,
@@ -57,6 +58,10 @@ class DashboardView(TemplateView):
                 order.event = re_dict[order.event_id]
                 order.event.order = order
 
+            # Exclude registered events
+            upcoming_events_interest = upcoming_events_interest.exclude(pk__in=registered_events_qs)
+            upcoming_events = upcoming_events.exclude(pk__in=registered_events_qs)
+
             # Past events is things you at one point paid for.
             # So you've paid for something, even if it was later refunded.
             past_events = Event.objects.filter(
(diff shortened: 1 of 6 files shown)


harrislapiroff:
Don't ask me why the june event starts today.

brambling/views/core.py
```diff
@@ -57,6 +58,10 @@ def get_context_data(self):
                 order.event = re_dict[order.event_id]
                 order.event.order = order
 
+            # Exclude registered events
+            upcoming_events_interest = upcoming_events_interest.exclude(pk__in=registered_events_qs)
+            upcoming_events = upcoming_events.exclude(pk__in=registered_events_qs)
```
melinath: should this use the list version instead, so that it doesn't use a subquery, or do we not need to worry about that because Postgres?

harrislapiroff: will the list version work? I wasn't sure and trust your judgment.

melinath:
I very much like the buttons, but I am concerned that this is a confusing path to go down... I really hate lists that look like they're ordered by date but which have some exceptions. Like... facebook's event lists are TERRIBLE. This is better because there's an explanation, but... it's still weird.

brambling/views/core.py
melinath: the list version... should work? If it doesn't it'll be real obvious real fast. we'd really need to do tests to know for sure which is faster, and at our current scale it probably doesn't matter.

harrislapiroff:
Maybe if there was a divider between the registered events and the upcoming? I think that would be clear but not overbearing.

harrislapiroff:
I also wouldn't mind integrating events you're attending into the upcoming events feed in order, but if we do that we should provide an obvious way to see all events you are registered for.

melinath:
Maybe a tab view? I'd also be down with a separate page showing registered events, maybe a giant button in the sidebar to take you there. That could also be a good place to put past events instead of the sidebar.

harrislapiroff:
Yeah, a tab view would be great! Will it be extremely hard to annotate the querysets with a registration status attribute? The current template logic is a little brittle...

Or maybe a template tag/filter that accepts an event and a user and returns details about the relationship between them.

melinath:
So I guess I'm leaning towards a separate page that has all your registered events (past, current, future) and leaving the dashboard for browsing. Then "Your Events" would need to point to the new page and we'd want a separate link for browsing upcoming events.

harrislapiroff:
If I add a divider can this be an acceptable intermediate step or will that be insufficient? I think a separate page is going to require a new view, a navigation restructure, and getting some annotations on querysets for registration status.

harrislapiroff:
(I think I prefer the divider to the heading because if they still look more or less like one list, it explains why each item only appears once, and I didn't like items sometimes appearing twice. We can't do the combined ordered list easily yet because the conditional button structure only works on the events that have order information attached.)

melinath:
I'm fine with just a divider for now, if you think that's a good step. I can take a look at making a separate page / annotated queryset later, if you think that idea makes sense?

harrislapiroff:
I think querysets annotated with user-specific data is going to have a lot of uses, including combined list so, definitely +1 on that. I'm interested in a separate page for your own events, but we'll have to figure out the UI so that it doesn't look silly if you have 0 or 1 events and also so that it's conveniently available. Tentatively +1 on that but I wanna discuss the UX.

harrislapiroff:
Also, I added a divider that's sort of a compromise between a divider and a heading:

screenshot 2015-04-30 15 17 06

melinath:
I like! And @nmorduch is also good with it. Merging.

Issue #479, opened 2015-04-28
Allow users to register for events without confirming their email address


melinath:
There's a weird contradiction where you can sign up without an account just fine, but if you want to sign up *with* an account you have to confirm your email address. This can actually lead to a situation where a person starts registering, then creates an account. We smartly auto-transfer their order to their new account - and they can't edit their cart because they haven't confirmed their address.

I think we should do the following:

* [x] Remove the requirement to confirm your email address before registering for an event
* [x] Auto-confirm addresses that are added via invitation
* [ ] Use confirmation as the metric for letting people *claim* old orders. (See #246 – wouldn't actually need to be completed as part of this ticket.)

melinath:
Also related: #362.

Issue #246, opened 2014-10-28
Allow users to claim their previous orders upon creating an account.


harrislapiroff:
Someone should be able to find their registration records on the site regardless of whether they originally registered with a Dancerfly account or can find the original receipt. I haven't actually looked at the new registration system yet, so I don't know whether or not this is already feasible. Just making a ticket so I don't forget.

melinath:
Then this is (I believe) a duplicate of #23.

melinath:
Closing as a duplicate.

melinath:
Re-openint because it's not. :-p

melinath:
I think it would make the most sense to handle this by having a view where folks can opt to take on old orders. Confirming your email would take you there, but we'd need another path as well, in case someone skipped it at first but wanted to change their mind later.

PR #486, opened 2015-04-29
Email confirmation improvements

### LINKED ISSUE #362: Email confirmation page 404 should have clearer message
It would probably be helpful to say something like, "This email has either already been confirmed or this confirmation link is invalid." Or maybe just "This email has already been confirmed." if that's possible and not a security concern.

melinath:
Some minor cleanup / improvements to email confirmation behavior. Resolved #362 and #479. Users can now register without confirming their email address, accepting an invitation confirms your address, and invalid tokens give you an informative message & resend link rather than a 404 error.

#246 (claiming old orders) is more involved, so I figured it would be best to keep as a separate ticket.


harrislapiroff:
We should probably append a bit to all emails to unconfirmed accounts that says something like, "If you did not register for Dancerfly or believe this message was not meant for you, please email hello@dancerfly.com." I'll open a ticket.

Issue #346, opened 2015-02-14
Reconsider stripe form display to have more labels


melinath:
This is an accessibility thing (see [bootstrap's note on labels](http://getbootstrap.com/css/#forms-inline)) and also apparently people are having trouble knowing what the mm/yy fields are for. Adding a label just for those fields looks really weird, so we need to reconsider the design in general.

melinath:
Dancerfly user I worked with today didn't understand what MM and YY meant, and was a little confused about the CVC number as well. (Hence why folks have those explanatory pictures usually, I guess!)

PR #487, opened 2015-05-01
Stripe form updates

melinath:
This makes some updates to the stripe form (see #346) and also fixes some serious bugs around credit card handling, including some that could cause data loss.


melinath:
I think the saved card payment form still needs to be updated to use cards with is_saved=True. And preferably also to use radio buttons plus the more compact layout used on the profile form changes.

PR #502, opened 2015-05-06
Saved reports

### LINKED ISSUE #465: Saved Reports
Event creators will often want to look at the same set of data over and over. Right now, that can mean going in and having to check/uncheck 10-15 boxes and apply certain sorts and filters. We should add a feature where you can save a view of the data and come back to it later - I envision a simple "Save this report" button next to the sort/filter options.

melinath:
Resolved #465 (saved reports). Provided a simple way for organizers to save a set of search/filter/column settings to return to at a later date.

The UI on this could use a slight cleanup... I wasn't sure whether I set up the bootstrap wrong or if we need to override something. (Or *not* override something.)

We also need a way to *delete* saved reports, but that is less crucial and requires some more UI work, so I was thinking it might be good to do as a separate ticket? But I could also do it as part of this ticket. We should also add this to the features list at some point.


melinath:
Though as part of this PR we should provide a way to *update* a report, for sure.

melinath:
Removing a report was pretty easy to add, so I went that route. I think that with that in place, we don't need an additional method to edit reports, at least for now.

harrislapiroff:
The whole table browsing UI requires a larger UI overhaul, which I'm not going to get too much into right now. Some thoughts though, so I remember to create a ticket:

Currently the controls for what is a single form is spread across multiple locations such that it doesn't look like a single form. This creates confusing and unintuitive behaviors. For instance, if I tap "Change columns and filters" make changes, then press cancel, those changes will take if I hit "Download CSV" or submit a search term, which is completely unexpected. This problem isn't new to this branch, but it's exacerbated by loading more controls into the same space. For instance, it would be nice and more intuitive if one could actually see the filters and columns that I am saving when I press the save report button. (This makes me think that all of these controls should be tucked into a sidebar that can be collapsed, similar to the grappelli admin, or integrated into an interactive javascript table, similar to a Google Sheets filter view.)

Things that probably should be fixed on this branch:

* [ ] I should not be allowed to create a new saved report identical to a previous saved report. The form to create a new saved report probably should not event be shown.
* [ ] The &times; in the dropdown menu should be a trashcan. &times; could imply unselecting rather than deleting.
* [ ] There should be a way to unselect the current report (i.e., go back to default filters and columns) (related to #503, so maybe worth saving for that)
* [ ] Floppy disks are silly as save icons. This doesn't even really need to be a save icon. A check or an "OK" or similar would probably suffice.

harrislapiroff:
It's also particularly unclear right now (partially because the filters/columns are visually separate from the saved reports form) what a saved report is.

harrislapiroff:
One other thought that might make the UI here better: Maybe this is a popover and not a dropdown?

harrislapiroff:
Not perfect, but I think this is better:

screenshot 2015-05-19 13 55 00
screenshot 2015-05-19 13 54 52

harrislapiroff:
I think this still allows you to create identical reports if you manually select all the same filters, but it at least doesn't display the form if you've got a report selected. Bouncing back to you @melinath.

harrislapiroff:
(I made some deeper SASS changes as well, but they were all things I've been pondering for a while and will pave the way for other UI cleanup ideas I have: better list groups in modals, a less aggressively colorful UI overall, et al.)

harrislapiroff:
Okay, now I'm done for real.

Issue #484, opened 2015-04-29
Thunderbird marking emails as spam


harrislapiroff:
Thunderbird apparently distrusts emails that contain URL links that go to a different location than the listed URL. It looks like this bug [dating back to 2006](https://blog.mailchimp.com/new-security-feature-in-thunderbird-triggered-by-click-tracking/) might be the culprit.

A few possible courses of action:

* Don't use Mandrill. (Not preferable. Mandrill is way easier to debug than our own SMTP server! Also we can do stuff like pause emailing and track click throughs!)
* Don't include an explicit URL at all in the HTML version of the email. (This may entail custom writing a non-HTML version of the email?)
* Include the explicit URL, but don't link it. (Is this possible or does it have to get autolinked somewhere along the way?)

melinath:
We have to link the URL. Some clients autolink, but others don't – that's the issue we were running into with Manchester Lindy.

I think it could fix it if we just had the link be text – "Click here to verify your account" – but that's assuming the mismatch was the issue.

melinath:
Hmm, that bug report seems to indicate that wouldn't be a fix, and that there is no fix, and that we should just make our emails prettier.

harrislapiroff:
I've found a few references to the URL mismatch feature of Thunderbird when searching for "Mandrill spam thunderbird" so I think there's a strong possibility that that's the issue.

I was thinking we link the email in text _and_ provide a "If the link above doesn't work, copy and paste this into your URL bar," sort of thing that emails often do.

But also, it's 2015. Do we still need to do that? Can we just link some text? The only issue I see with that would be that we should provide a plain-text alternative.

harrislapiroff:
Why wouldn't that be a fix? If the issue is a mismatch between the text of a link and the actual link, we could fix it by not having an actual link, no?

harrislapiroff:
> That means if you send an innocent email newsletter to your subscribers with a complete URL spelled out in the email (for example, "Please visit us at http://www.mysite.com" ) and then you enable click tracking for the email, that link would set off the scam alert in Thunderbird. 

So we could avoid that by either not spelling out the URL or not having an HTML link on the spelled out URL. Right?

melinath:
The bug report you linked indicates that it's an issue for text links as well.

>Scammers do this all the time. They’ll place a link in an email, such as "Verify Your eBay Password" but if you click it, you get taken to some evil website, where they steal your password, or download spyware onto your computer.

[...]

Issue #402, opened 2015-03-06
Display multiple transactions on shopping summary page


nmorduch:
The receipt-like list of things bought on the order should be split up so that there is one list for each transaction. There should, if possible, also be something saying whether payment for each transaction has been received.

harrislapiroff:
It might even be nice to have receipt detail pages for each transaction or at least a tabbed interface. Maybe one of the tabs can be overall summary and display the sort of combined data we currently display.

nmorduch:
I was thinking that would get rid of the nice overall summary but I think what you're suggesting, having a tab for the summary, is a good solution. I was thinking of having a nav on the left like in attendee assignment

harrislapiroff:
:+1:

harrislapiroff:
Another way of thinking about this ticket might be that there should be a distinction between your transaction summary (records of what you purchased, when, how the financial transaction was carried out, and whether it was refunded) and your inventory (what you currently own). We might even want a new (separate) inventory view if it isn't combined into this view.

harrislapiroff:
Although that's actually a separate ticket, I suppose. Just spitballing.

melinath:
The order summary view should absolutely have a list of transactions and a list of inventory (including attendee assignment) side-by-side. I think it would make sense to have a separate view for details on each transaction, plus linking to the item assignment / data editing views.

melinath:
Resolved with merge of #504

Issue #398, opened 2015-03-06
Better icon for refund, with help text


nmorduch:
http://fortawesome.github.io/Font-Awesome/icon/undo/ looks better than the refreshy option in the list 

![undo](https://cloud.githubusercontent.com/assets/1051450/6530630/7872c540-c400-11e4-8a32-19cc98cf94f4.png)

And there should be text next to it because no icon is going to be clear.

melinath:
Resolved with merge of receipt redesign.

Issue #393, opened 2015-03-06
Use parentheses for negative in order summary


nmorduch:
Fees of "-$5" is confusing. Especially when it's "$-5" :p
But it is good to note that in the sum it is counted as negative.

melinath:
We currently handle this formatting with a template tag in zenaida. That could hypothetically move to brambling... it was in zenaida because of argus, IIRC

melinath:
Resolved with merge of #504

PR #504, opened 2015-05-07
Receipt redesign

melinath:
I've tried to clean up and clarify receipts / order attendees. I believe this branch will close at least the following tickets:

* [x] #287 Better interface for marking checks as paid.
* [x] #398 Better icon for refund, with help text
* [x] #402 Display multiple transactions on shopping summary page 
* [x] #393 Use parentheses for negative in order summary 
* [x] #290 Refunded items display in order summary 
* [x] #326 Payment form UI
* [x] Emails
* [x] #505 Only apply discounts to current cart.
* [x] Discounts (do they actually get calculated right?)

This isn't quite done. The receipt views still need another quick pass, and this needs to be integrated into emails (possibly touching on #484 Thunderbird marking emails as spam).


melinath:
Attendees whose passes have been refunded will not be accessible to edit from the summary – but that will be resolved by the branch which removes item categories.

melinath:
@harrislapiroff This is ready for review. I think you'll like it? I hope!

nmorduch:
Can these not be buttons? It's confusing that the change on hover but aren't clickable. Maybe labels instead?

nmorduch:
Ohhhhh because it's the same design for attendees as organizers. I think that should be changed though, before merging.

nmorduch:
What is the " Better interface for marking checks as paid."?

melinath:
It would be pretty trivial to switch those to labels for attendees - that template block is overridden completely on the organizer side. Feel free to make that change!

Better interface is having actual buttons instead of organizers having to know they can click on what looks like a piece of text. I'm sure it could be improved further, though. Maybe a tooltip?

nmorduch:
Tooltips are pretty eh for touch screen users :/

melinath:
The organizer interface is not really intended to be optimized for mobile.

melinath:
See also: the table views.

nmorduch:
Eventually I'd live everything to work for all devices though? e.g. ipads are quite reasonable

melinath:
Ah, fair enough. Switching to buttons is an improvement and all I want to do in this PR. If we want to do more later we can just keep the other ticket open.

nmorduch:
:+1:

harrislapiroff:
Agree with discussion here. FWIW, IMO better interface in #287 means be not having to go to a record's detail view to mark a check as paid, since organizers are likely to be doing that in bulk. But that conversation already happened on the ticket.

harrislapiroff:
This looks nice, but for a future redesign, I think we've outgrown sharing the templates between the order view from the attendee side and the order view from the organizer side. The purpose of each view is different. For the organizer it should be as efficient/compact as possible and can rely on the organizer's familiarity with the interface over time since they'll be seeing it a lot. For the attendee it should be more verbose, as it is in this redesign.

harrislapiroff:
NICE WORK ON THE PAYMENT ACCORDION

harrislapiroff:
Although, maybe in a future ticket, we should mimic other checkout processes such that if you have a saved card, it just displays that and then provides a nearby button/link to use a different card and swap out the form. I think it'd be nice if credit card was a single panel.

But seriously it looks good.

harrislapiroff:
Agree with Naomi that unclickable buttons are extremely confusing. It feels like I'm supposed to click that to confirm my registration.

harrislapiroff:
If you address those things above, either by polishing the UI now or creating future tickets, I think this is otherwise good to go, @melinath.

melinath:
1. Oh right, that is what we'd talked about with the better interface for check confirmation. Whoops! Still, I think this is an improvement...

2. +1 on splitting the templates for receipts. I almost did that here, but didn't relish the idea of revamping that interface completely on top of everything else...

3. yay I'm glad you like it!

The only thing I would want to maybe not open a new ticket for is the unclickable button issue. I guess a quick fix would be to make the button disabled on the customer side of things...? Slash, do you think it's safe to merge this without fixing that and then get it in a later ticket?

melinath:
I don't know how long we'd want to leave that.

harrislapiroff:
I think disabled buttons would still be confusing since those still have a different cursor. Maybe just fix it here if it's quick though? Can you use an `{% if %}` and just swap it to a message or label or [colored text](http://getbootstrap.com/css/#helper-classes-colors)?

melinath:
screen shot 2015-05-14 at 11 47 54 am

harrislapiroff:
Looks good. I'm happy for you to merge if you're happy with it!

Issue #497, opened 2015-05-06
Add unwanted email support text to messages to unconfirmed email addresses


harrislapiroff:
Since it's possible to perform actions and register for events on Dancerfly without a confirmed email address, it's possible (though admittedly silly) for someone to register for events with someone else's email address. Just to cover our bases, we should probably append some text to all messages that go to unconfirmed email addresses saying something like:

> If you did not register for Dancerfly or believe this message was intended for someone else, please email hello@littleweaverweb.com.

Issue #448, opened 2015-04-03
Address for check mailing in registration email doesn't include line breaks.


harrislapiroff:
Mine looks like this:

![screenshot 2015-04-03 13 27 03](https://cloud.githubusercontent.com/assets/296631/6985814/2ef2427c-da05-11e4-87d8-d9657f000c83.png)

melinath:
That's odd... the `<br />` tags are definitely in the template, so they're probably being rendered... I guess something somewhere along the way is stripping them out?

chigby:
@harrislapiroff Do you know if this is still happening? The email templates have undergone fairly significant changes since this issue was created.

harrislapiroff:
I have no idea, honestly :\. Maybe we should have some views for sample renderings of emails?

melinath:
We do have views for sample renderings of emails ;-)

harrislapiroff:
jklol

melinath:
It looks like this is resolved, looks like as part of #506.

![screen shot 2016-01-06 at 2 59 43 pm](https://cloud.githubusercontent.com/assets/299979/12157358/2fa45e24-b486-11e5-8b01-b4d7940fa4f7.png)

![screen shot 2016-01-06 at 3 00 31 pm](https://cloud.githubusercontent.com/assets/299979/12157374/49bbdd3c-b486-11e5-92d2-7f685f03bb85.png)

PR #506, opened 2015-05-07
Better emails

melinath:
This branch will introduce some email improvements - in particular, the ability to preview emails in the browser, and a base [Zurb Ink](http://zurb.com/ink) template. I don't know if I'll redo all the templates to use it or just do a proof-of-concept...

Things this could touch on:

* [x] #484 Thunderbird marking emails as spam
* [x] #497 Add unwanted email support text to messages to unconfirmed email addresses 
* [x] #448 Address for check mailing in registration email doesn't include line breaks.


melinath:
Receipt emails still need to be updated... I'd be fine leaving those until after merge. I haven't updated the home invite template either, because I'd rather just do #515.

melinath:
Okay, this is ready for review. Note: After #504 lands, the email templates will need to be updated to account for the new summary_data structure.

harrislapiroff:
The templates need a bit of work (the footer image isn't scaling down, I think). You can preview templates using our Mandrill dashboard: https://mandrillapp.com/templates/code?id=test-template

The dropdown menu in the corner can fake a preview in-browser or send an actual email using the html in the template.

harrislapiroff:
I don't have a good way of thoroughly checking all the python, but the few email-triggering things I attempted all seemed to send emails OK.

melinath:
@harris did you catch that you can navigate to `/mail/confirmation/` and similar to preview / send emails?

harrislapiroff:
oop I missed that

melinath:
I think this is ready to merge now. Footer image is just the size of the biggest column, which I think is fine.

PR #507, opened 2015-05-07
Remove item category

melinath:
This is the PR for removing the Item.category field! See also: the lengthy discussion on 34f5f7347ae57cbe7a2591578e4bc6d25ce498cc.

Remaining (known) issues:

* [x] Model tables – pass status and pass type no longer exist; what effect does that have?


melinath:
Okay, I think this is ready.

harrislapiroff:
Attendee creation form title needs to be removed or revised:

screenshot 2015-05-14 15 11 06

harrislapiroff:
This seems to work more or less as expected, less the UI change above.

I want to reiterate my reservations about this PR. I still feel like it's a mistake to give us no way to determine how many attendees should correspond to a shopping cart in terms of UX and future features, but I'll concede that the base case of purchasing a single registration for a single attendee seems to work smoothly. With the UI above fixed, this is ready to merge.

melinath:
@harrislapiroff Fixed the UI issue. I guess I'll merge now?

harrislapiroff:
:+1:

PR #519, opened 2015-05-10
Removed references to Invite.HOME and code which supported multiple housemates existing.

### LINKED ISSUE #515: Remove "housemate" managing
Ultimately, we want to let folks have multiple attendees / homes per account. We don't want to let people share their house information with their housemates. While well-intentioned and potentially cool, this is ultimately not *that* useful and adds complexity we could easily do without.

melinath:
Resolved #515 Remove "housemate" managing. This PR represents the first step - removing the ability of people to add other folks to their home. This makes the Person-Home relationship essentially a one-to-one relationship for now.

Later the idea is to set people up to have multiple saved configurations for attendees and homes. For now, I just want to avoid someone actually managing to set up a people-home situation (since it would make migrations more complicated later.)


harrislapiroff:
This is missing a migration—but it's one that I don't think has any effect on the database, so I'll let you figure out how you want to slip that migration in.

melinath:
yeah... I think I left it out because I wanted to make sure we didn't get another fucking merge migration if we could avoid it.

harrislapiroff:
word

harrislapiroff:
I'm not sure why, but saving the home editing form is not working for me. It just blanks and does not seem to save. Let me know if you can't reproduce.

melinath:
I can reproduce. It happens if the user is new & doesn't have a home attached to them yet.

melinath:
Fixed! Also, I checked, and this is handled correctly on the hosting form.

harrislapiroff:
Seems dandy. Merging.

PR #520, opened 2015-05-12
Added links to remote payment detail pages from finances view.

### LINKED ISSUE #508: Link to remote payment detail pages from transaction list
From FDW:

> As icing, if the Stripe API gives you easy access to the Stripe payment ID, it might be pleasant to include a link to the Stripe dashboard page for that payment in your Finances view: https://dashboard.stripe.com/payments/{{ STRIPE_ID }}

I think this would be pretty easy to do, though not necessarily 100% future-proof... For Dwolla, the URL (seems to be) https://dwolla.com/activity#/detail/{{ REMOTE_ID }}.

melinath:
Resolved #508.

I've tested this; the links seem to work properly for stripe vs. dwolla, test vs. live, and purchase vs. refund.

### DIFF
--- a/brambling/templates/brambling/event/organizer/finances.html
+++ b/brambling/templates/brambling/event/organizer/finances.html
@@ -21,6 +21,7 @@
                     <th>Amount</th>
                     <th>Dancerfly Fee</th>
                     <th>Payment Fee</th>
+                    <th></th>
                 </tr>
             </thead>
             <tbody>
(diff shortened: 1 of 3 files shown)


melinath:
The change is pretty simple, testing was annoying. Mainly doing as a PR in case there are design concerns...

brambling/templates/brambling/event/organizer/finances.html
```diff
@@ -35,6 +36,13 @@
                         <td>{{ transaction.amount|format_money:event.currency }}</td>
                         <td>{{ transaction.application_fee|format_money:event.currency }}</td>
                         <td>{{ transaction.processing_fee|format_money:event.currency }}</td>
+                        <td>
+                            {% if transaction.get_remote_url %}
+                                <a class='btn btn-default' href="{{ transaction.get_remote_url }}" target='_blank'>
+                                    View on {{ transaction.get_method_display }}
+                                    <i class='fa fa-fw fa-external-link'></i>
+                                </a>
+                            {% endif %}
```
harrislapiroff: `</td>`?

melinath: :-p

harrislapiroff:
This looks good. It might be good at a later point to also have this data propagate down to the orders table to reduce cross referencing that the organizer needs to do.

Issue #509, opened 2015-05-07
Send metadata to Stripe and Dwolla with charges


melinath:
Stripe and Dwolla both allow us to send notes and/or metadata along with charges, which can help organizations keep track of charges in the remote service's dashboard. We should take advantage of this.

This was requested by FDW along with #508.

melinath:
Fixed with the merge of #531.

PR #531, opened 2015-05-26
Remote metadata

melinath:
This PR starts sending metadata along with all stripe and dwolla payments and refunds, as well as adding metadata to past stripe payments. (Past Dwolla payments can't be updated because their API doesn't allow it.)


melinath:
This PR resolves #509.

harrislapiroff:
Looks good.

PR #534, opened 2015-05-28
Boughtitem value caching

### LINKED ISSUE #337: Cache most important values on BoughtItem at buy-time
We should cache certain values (item name, itemoption name, amount paid) on BoughtItem in addition to linking to the correct itemoption. This will save some queries, and it will also help us when we want to implement sliding scales (#5).

melinath:
This PR resolves #337. It sets up caching of important values on BoughtItem and BoughtItemDiscount at creation time. This has a double benefit of reducing required queries / query complexity related to calculating prices and discounts AND ALSO preventing unexpected behavior if organizers change a price / discount amount or delete an item or discount altogether.

To look into:

* [ ] Discount bug
* [ ] Failing/Erroring tests

### DIFF
--- a/brambling/admin/admin.py
+++ b/brambling/admin/admin.py
@@ -1,5 +1,6 @@
 from django.contrib import admin
 from django.contrib.auth.admin import UserAdmin
+from django.db.models import Count
 from brambling.models import (Person, Event, DanceStyle,
                               EnvironmentalFactor, DietaryRestriction,
                               HousingCategory, CustomForm, CustomFormField,
@@ -128,8 +129,24 @@ class OrganizationAdmin(admin.ModelAdmin):
     )
     raw_id_fields = ('owner',)
     filter_horizontal = ("dance_styles", "default_event_dance_styles", "editors")
-    list_display = ('name', 'created')
+    list_display = ('name', 'created', 'default_application_fee_percent', 'published_events', 'all_events')
     ordering = ('-created',)
+
+    def get_queryset(self, request):
+        qs = super(OrganizationAdmin, self).get_queryset(request)
+        return qs.annotate(all_events=Count('event')).extra(select={
+            'published_events': """
+            SELECT COUNT(*) FROM brambling_event WHERE
+            brambling_event.organization_id = brambling_organization.id AND
+            brambling_event.is_published = 1
+            """
+        })
+
+    def all_events(self, obj):
+        return obj.all_events
+
+    def published_events(self, obj):
+        return obj.published_events
 
 
 class EventAdmin(admin.ModelAdmin):
@@ -178,8 +195,13 @@ class EventAdmin(admin.ModelAdmin):
     raw_id_fields = ('organization',)
     filter_horizontal = ("dance_styles", "additional_editors")
     radio_fields = {'api_type': admin.HORIZONTAL}
-    list_display = ('name', 'created')
+    list_display = ('name', 'organization', 'is_published', 'is_frozen', 'created', 'application_fee_percent')
+    list_filter = ('organization', 'is_published', 'is_frozen')
     ordering = ('-created',)
+
+    def get_queryset(self, request):
+        qs = super(EventAdmin, self).get_queryset(request)
+        return qs.select_related('organization')
 
 
 class CustomFormFieldInline(admin.TabularInline):
--- a/brambling/forms/organizer.py
+++ b/brambling/forms/organizer.py
@@ -194,6 +194,11 @@ class EventForm(forms.ModelForm):
         if not self.organization_editable_by:
             del self.fields['editors']
 
+        if self.instance.pk is None:
+            self.instance.application_fee_percent = organization.default_application_fee_percent
+
+        # Always display the timezone that is currently chosen,
+        # even if it wouldn't otherwise be displayed.
         if self.instance.pk is not None:
             timezone = self.instance.timezone
             if (timezone, timezone) not in self.fields['timezone'].choices:
--- a/brambling/migrations/0011_auto_20150528_1806.py
+++ b/brambling/migrations/0011_auto_20150528_1806.py
@@ -1,0 +1,46 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+import django.db.models.deletion
+import django.core.validators
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('brambling', '0010_auto_20150528_1805'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='boughtitem',
+            name='item_description',
+            field=models.TextField(blank=True),
+            preserve_default=True,
+        ),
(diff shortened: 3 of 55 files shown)


melinath:
I haven't yet had a chance to do a QA pass on this, so it may have some bugs. But I *think* I got all the references and handled the conversion correctly. Bits I know I touched:

* modal cart (order base template)
* _attendee_summary.html
* _transaction_receipt.html
* attendees.html
* shop.html (sidebar cart)
* order_alert email (body & body_inlined & body_plaintext)
* order_receipt email (body & body_inlined & body_plaintext)
* summary data calculations (price, savings)

Remaining concerns:

* Not 100% sure what effect this will have on model tables, in terms of being able to filter by item options.
* Alternate solution to this problem would be to disallow editing of discounts / item options that have already been purchased, and to remove the ability to ever *really* delete them. That *may* be a better route – really not sure – but it doesn't necessarily conflict with this one. Caching would still be useful from a query complexity perspective.

melinath:
After talking about this with someone, I am once again convinced that caching is the future!

harrislapiroff:
I'll look over this branch, but as you said, it's not mutually exclusive with keeping items around forever. I think we should (eventually) not delete items but either offer an exposed interface to permanently disable them or keep them in the database for filtering/data purposes.

melinath:
So just to clarify, the effect on model tables is that you won't be able to filter for item options that you've deleted. We could "fix" that by filtering by item option name & item name, but... that isn't much of a fix and could easily result in incorrect groupings.

harrislapiroff:
Yeah, that's more or less what I was thinking in terms of keeping deleted items in the database (though I guess that would also entail keeping deleted item options in the database...)

With those concerns in mind, do you still want this reviewed and merged?

melinath:
Oh, yes, absolutely. If we do end up keeping deleted items around, we can always re-create any deleted items from the cached values!

melinath:
Plus this ticket being merged will help with sliding scale implementation.

melinath:
and Manual CRUD.

melinath:
This will need a merge migration, or just switching the dependency of the 0011 migration to the latest master migration.

harrislapiroff:
Not sure if it's this branch or master, but it seems to fail some tests:
```pytb
Creating test database for alias 'default'...
.F.E.................E..................
======================================================================
ERROR: test_dwolla_payment_form (brambling.tests.functional.test_order_forms.PaymentFormTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/harris/.virtualenvs/brambling/lib/python2.7/site-packages/mock.py", line 1201, in patched
    return func(*args, **keywargs)
  File "/Users/harris/Projects/Clients/LittleWeaver/django-brambling/brambling/tests/functional/test_order_forms.py", line 191, in test_dwolla_payment_form
    form = DwollaPaymentForm(order=order, amount=Decimal('42.15'), data={'dwolla_pin': pin, 'source': source}, user=person)
  File "/Users/harris/Projects/Clients/LittleWeaver/django-brambling/brambling/forms/orders.py", line 550, in __init__
    self.sources = dwolla_get_sources(dwolla_obj, event)
  File "/Users/harris/Projects/Clients/LittleWeaver/django-brambling/brambling/utils/payment.py", line 120, in dwolla_get_sources
    access_token = dwolla_get_token(user_or_order, event.api_type)
  File "/Users/harris/Projects/Clients/LittleWeaver/django-brambling/brambling/utils/payment.py", line 74, in dwolla_get_token
    dwolla_set_tokens(dwolla_obj, api_type, oauth_data)
  File "/Users/harris/Projects/Clients/LittleWeaver/django-brambling/brambling/utils/payment.py", line 35, in dwolla_set_tokens
    expires = timezone.now() + datetime.timedelta(seconds=data['expires_in'])
KeyError: 'expires_in'

======================================================================
ERROR: test_dwolla_charge__user (brambling.tests.integration.test_dwolla.DwollaChargeTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/harris/Projects/Clients/LittleWeaver/django-brambling/brambling/tests/integration/test_dwolla.py", line 55, in test_dwolla_charge__user
    source='Balance',
  File "/Users/harris/Projects/Clients/LittleWeaver/django-brambling/brambling/utils/payment.py", line 139, in dwolla_charge
    access_token = dwolla_get_token(sender, event.api_type)
  File "/Users/harris/Projects/Clients/LittleWeaver/django-brambling/brambling/utils/payment.py", line 74, in dwolla_get_token
    dwolla_set_tokens(dwolla_obj, api_type, oauth_data)
  File "/Users/harris/Projects/Clients/LittleWeaver/django-brambling/brambling/utils/payment.py", line 35, in dwolla_set_tokens
    expires = timezone.now() + datetime.timedelta(seconds=data['expires_in'])
KeyError: 'expires_in'

======================================================================
FAIL: test_subject__organization_editor (brambling.tests.functional.test_invites.InviteTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/harris/Projects/Clients/LittleWeaver/django-brambling/brambling/tests/functional/test_invites.py", line 25, in test_subject__organization_editor
    self.assertEqual(mail.outbox[0].subject, "{} has invited you to edit {}".format(invite.user.get_full_name(), event.name))
AssertionError: u'User4 Test has invited you to help manage Test event' != 'User4 Test has invited you to edit Test event'

----------------------------------------------------------------------
Ran 40 tests in 13.458s

FAILED (failures=1, errors=2)
Destroying test database for alias 'default'...
```

harrislapiroff:
Again, not sure if this is on this branch or on master, but I attempted to create a discount and apply it to a previous order (from the organizer side), but the discount had no effect.

harrislapiroff:
With those above things fixed (or determined to be on master) _and_ making sure that interactions between deleted items and discounts is reasonable, I think that this is ready.

harrislapiroff:
I also still think at some point we should keep all items in the database to refer to later. But the caching is good anyway, since it preserves the state of the item at the time of purchase, which is also important.

melinath:
Not applying discounts to completed orders is the new correct behavior, because it was really weird and unintuitive that you could do that in the first place. :-p

harrislapiroff:
Oh. That makes sense. We should probably display a message or something saying it's not allowed—or not show the tab at all. But that sounds like an issue on master.

melinath:
Merging in master fixed the invite bug. The other two are also master-related. I am not 100% sure, but I'm fairly confident that they are the result of the dwolla test refresh token expiring...

melinath:
So I think they're safe to ignore for the purposes of merging this branch.

harrislapiroff:
Cool. I'll do a little more checking things, but I anticipate merging this in a few minutes.

harrislapiroff:
Looks good. It's a bit weird that you can end up with a discount that has no applicable items, but I can't think of a better way of handling that and it doesn't seem severe, since the worst case is that an organizer has a discount that doesn't do anything.

PR #541, opened 2015-06-02
Django REST framework (maybe ready?)

melinath:
This pull request represents initial steps towards using Django REST Framework + Ractive.js to have a dynamic shopping experience. Woo!

This has been kinda tricky because (some of) the actions we want to take don't really map well to the standard REST CRUD layout and I'm not 100% sure I made all the right calls. But it basically works.

Outstanding things that come to mind:

* [ ] Resizing images via the Daguerre API.
* [ ] Displaying step bar.
* [ ] Formatting date display correctly (i.e. adding helper function to humanize dates)
* [x] Cart expiration setting & display

Not sure how finished you wanted to get this before merging. Currently, this provides a separate page (at ``/org-slug/event-slug/order/ractive-shop/``) which is the ractive-powered shop, while leaving the normal shop untouched, if you want to check it out.

### DIFF
--- a/brambling/admin/admin.py
+++ b/brambling/admin/admin.py
@@ -1,5 +1,6 @@
 from django.contrib import admin
 from django.contrib.auth.admin import UserAdmin
+from django.db.models import Count
 from brambling.models import (Person, Event, DanceStyle,
                               EnvironmentalFactor, DietaryRestriction,
                               HousingCategory, CustomForm, CustomFormField,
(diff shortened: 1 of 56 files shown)


melinath:
This is actually working pretty darn well... but we're still missing the three items listed above. Once those are completed, this should be ready to switch over... which I think I'd like to do sooner rather than later, since otherwise it'll get out of sync and we might miss something in the transition.

That being said, it would be good to sink a couple hours into writing tests for the api before merging, I think. I'd like to be able to replace the current shop with a fair degree of certainty that all hell won't break loose.

harrislapiroff:
This seems like a good start. I'm not sure everything is architected the way I would do it, but since it leaves the original store untouched I figured we could merge it and keep working on it there.

PR #542, opened 2015-06-03
Form improvements

melinath:
Adds support for choice fields (radio, select, checkbox multiple, and select multiple) and field help text. Bam.

### DIFF
--- a/brambling/templates/brambling/event/organizer/custom_form_form.html
+++ b/brambling/templates/brambling/event/organizer/custom_form_form.html
@@ -91,11 +91,11 @@
 
 			var setChoicesDisabled = function(ele) {
 				var $ele = $(ele),
-					$choices = $ele.parent().parent().find('textarea[id$=choices]');
+					$choices = $ele.parent().parent().find('textarea[id$=choices]').parent();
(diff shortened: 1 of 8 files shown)


brambling/templates/brambling/event/organizer/custom_form_form.html
```diff
@@ -87,6 +87,30 @@
 		(function () {
 			"use strict";
 
+			var choice_types = ['radio', 'select', 'checkboxes', 'select_multiple'];
+
+			var setChoicesDisabled = function(ele) {
+				var $ele = $(ele),
+					$choices = $ele.parent().parent().find('textarea[id$=choices]');
+				if (choice_types.indexOf($ele.val()) !== -1) {
+					$choices.prop('disabled', false);
```
harrislapiroff: Would it be okay to just hide and unhide the `.form-group` instead?

harrislapiroff:
This looks really great, but output in the organizer interface for multiple selects is imperfect:

screenshot 2015-06-03 15 18 05

melinath:
oh dear. Right.

PR #543, opened 2015-06-03
Fixes to Chosen on formsets

### LINKED ISSUE #406: Chosen + Formset = D:
When dynamically creating a new item in a formset, all the chosen's generated HTML seems to get duplicated, but the chosen behavior does not, basically rendering any dropdown selects on dynamically created items useless. :(

See this in action when creating new item options on the item editing view.

* [x] Item options
* [x] Form fields

harrislapiroff:
Resolves #406


melinath:
Works as advertised. :-)

melinath:
I like how switching to using ``<script>`` tags also made everything just generally cleaner.

PR #547, opened 2015-06-20
Add 'blues' as dance style.

j-po:
Eventually, it'd be nice for the dance style field to be free-form, but as a stopgap, I'd like to add "Blues".

### DIFF
--- a/brambling/models.py
+++ b/brambling/models.py
@@ -36,6 +36,7 @@ from brambling.utils.payment import (dwolla_refund, stripe_refund, LIVE,
 DEFAULT_DANCE_STYLES = (
     "Alt Blues",
     "Trad Blues",
+    "Blues",
     "Fusion",
     "Swing",
     "Balboa",


melinath:
Hey, Jonathan! What need would "blues" fill that isn't filled by the combination of " trad blues", "alt blues", and "fusion"?

melinath:
(We consciously chose not to include "blues" as an option in the beginning because it is, for better or worse, ambiguous to a lot of people. But you could potentially convince us to change our minds.)

j-po:
Ah, okay. 'Round these parts, we generally consider "Blues Dancing" to be dancing to blues music (of whatever era) and "Fusion Dancing" as a catch-all for everything else. The term "Trad Blues" is pretty much out of use here.

j-po:
(Optionally, see also:
- Damon on terminology and appropriation and stuff - http://damonstone.dance/articles/appreciation-versus-appropriation/
- PyCon talk on inclusive versus restrictive schemas. Schemata? For the more general case. https://www.youtube.com/watch?v=PYYfVqtcWQY)

melinath:
@j-po Aren't you in the San Francisco area? I thought that y'all still had Shades operating as a "blues" venue while playing fusion music... but whatever the case, the fact is that not everywhere in the country / world is as enlightened on the subject.

That being said, we try to hold ourselves to a higher standard. The trad vs alt blues terminology has been called out as appropriative more than once; no matter how "useful" it is, we shouldn't be promoting it. (Personally, I think that fusion has veered off far enough from blues that there's an in-between form that doesn't have a good name other than "alt blues" – but I don't think there are any events currently that actually use that / we can cross that bridge when we get there.)

I'm not accepting your pull request because changing the "default" styles won't change what's actually in the database. Our preferred fix would be to do the following:

1. Remove "Alt Blues" as a dance style.
2. Rename "Trad Blues" to "Blues".
3. Add a `description` field to dance styles.
4. Create a more elaborate dance style selection & display interface which makes clear to organizers & attendees what is meant by a given dance style.

In the future, we could also consider, say, allowing people to report events which are mislabeled. If we want to help events not appropriate by using "alt blues", we certainly can't ever make this a free-form field.

If you would like to contribute a PR that adds the description field and the relevant UI elements, that would be welcome; otherwise, I'm doing items 1 and 2 immediately and we'll get to the rest as we can.

j-po:
Thanks!

Shades operates as a mixed Fusion and Blues venue, though I expect this doesn't get communicated to newcomers as well as it could.

dlinet:
Hey Folks,
I realize this is an older article but I have a question if anyone is still listening. Has much of the dance community come to a consensus or thrown out a new idea of what "alt-blues" should be called if it isn't actually a blues form? I like that style of dance and would love to know how to find it in cities without seeming insensitive to the history of blues, but still wanting to find the style of dance that I like (previously called alt-blues?) .

harrislapiroff:
Hey, @dlinet! Short answer is that Dancerfly decided to remove the "alt blues" designation in a way similar to what @melinath listed above. Most events classified as "alt blues" could probably be better classified as "fusion".

Longer answer here: http://www.littleweaverweb.com/blog/2015/07/03/appropriation/

dlinet:
Hi @harrislapiroff,
Thanks for the further information! That longer answer makes sense to me as to why "alt-blues" was removed as a name. I'm still curious if any dancers of the previous "alt-blues" scene are thinking of a new name for the dance form. In my personal experience, "fusion" dances and "fusion" events have been quite different than previous "Alt-blues" dances and events. I'm not an expert on the topic but from my experiences there is a lot of latin and social dance influence in many of the fusion events I have attended, while the "alt'blues" events seem to draw more from modern dance, contact improvisation, blues idiom dance, and club dancing. Does anyone have experience with PUX or the Portland underground exchange? I am curious if they are suggesting the dance be called "underground" instead. I like the sound of that, but am not sure if that is something people/other dancers are thinking about.

melinath:
@dlinet This is still an ongoing discussion in the blues / fusion communities. In my experience, "fusion" and "alt-blues" events are the same. However, the fusion community contains two separate drives within itself – which I think is the same trend you're noticing. On the one hand, you have people who want to encourage dancing (and mastering) a variety of styles and then fusing them to music that also combines elements from the base musical styles belonging to those dances. And on the other hand, you have people who want to basically do "blues" but adapted to modern popular music – who see "fusion" as its own independent dance style.

Right now, both of those groups refer to themselves as "fusion" (or "alt-blues"). I think it would be useful to have a separate name for the individual dance style just so that people would acknowledge the existence of a dance style beyond just "fusing" whatever you happen to already know, but to my knowledge nobody's come up with a decent alternative.

FWIW here's how [PUX describes themselves](https://pux.open-dance.net/):

> Music at PUX will be the same as it has been previously, different music styles that lend themselves to different styles of dance, a good amount can be said to be Alt-Blues. Also, Fusion, depending on your definition, applies here.

Issue #332, opened 2015-02-10
User-to-user Ticket Transfers


melinath:
It's fairly common for attendees to (want to) transfer tickets to other attendees. A lot of events support this. Those that don't probably don't because it's a pain in the ass, not because they actually care.

It would be cool if we could implement a nice way of handling this. Essentially it would come down to a shareable link that transfers ownership from one user to another.

Though perhaps this is already covered well enough by letting people edit attendee data more easily?

melinath:
Nah, transferring ownership is more reliable.

melinath:
First / separately we should disallow implicit transfers for already-purchased tickets.

melinath:
I think we could pretty easily handle this via the Invite system... here's what I would propose:

For each item that's transferred, we create an Invite object that points to that object. When you accept the invite, it makes a copy of the BoughtItem, attached to the invited person's Order and priced at $0, and marks the original BoughtItem as Transferred.

melinath:
This was resolved with the merge of #554!

PR #554, opened 2015-07-11
Ticket transfers

melinath:
This is an incomplete implementation of ticket transfers (#332) which I am making into a pull request because I have forgotten what still needs to be done. I almost forgot the branch existed. o_O As I figure it out I will add notes here.

So. Things that need to work:

* [x] User can initiate a transfer of an item
* [x] Target can accept transfer (via the invite system)
* [x] Item is still recorded at price for former user
* [x] Item is also recorded for new user, but at $0
* [x] If an attendee has no more items, delete them
* [x] Admins can turn ticket transfers on and off

Things to create separate tickets for:

* [ ] User can cancel a transfer of an item
* [ ] Add tests of get_summary_data
* [ ] Freeze attendee info after checkout (i.e. no implicit transfers.) Link to transfer from attendee detail page.


melinath:
So I've looked over this, and I found one issue, but I think this is otherwise ready for merge, actually.

brambling/models.py
```diff
@@ -739,6 +727,126 @@ def get_icon(self):
...
+                    event=event,
+                    person=request.user,
+                )
+            except Order.DoesNotExist:
+                pass
+
+        # Next, check if there's a session-stored order. Assign it
+        # if the order hasn't checked out yet and the user is authenticated.
+        if order is None:
+            code = self._get_session_code(self.request, event)
```
chigby: I think `self.request` should just be `request`.

chigby:
Other than that one code comment, I'd say this is ready to merge.

brambling/models.py
melinath: Yes. Yes it should. :-/

melinath: Fixed plus tests!

melinath:
Ready for re-review!

PR #573, opened 2015-08-18
Allow exporting the Finances table as a CSV

chigby:
Puts a button the Finances page that exports what's on the table as a CSV. 

Implementation-wise, it's highly possible this could be more idiomatically Django. My strategy was to make a Python class that controls the formatting logic (which I copied from the template) and call it from the view. In theory this class could also be used to generate the data in the template -- with some tweaking (e.g. generating a link for the order number). 

Possible open questions:

 * Button positioning. It doesn't match the Attendees or Orders pages, but is possibly more obvious without all the rest of the controls. My technique here is called "don't mess with CSS until absolutely necessary."
 * I might be doing unnecessary work to get the timezone right in `format_timestamp`. The template uses a special function I couldn't really get to work in the view.
 * It worked for my trivial data set (1 transaction), but it might not for something larger. If there's an easy way for me to get a bunch of seed data, that could be handy. Or at least test the CSV code against something more closely resembling the actual data.

So! Hope it seems good.

### DIFF
--- a/brambling/tests/functional/test_finance_table.py
+++ b/brambling/tests/functional/test_finance_table.py
@@ -1,0 +1,57 @@
+from django.test import TestCase
+
+from brambling.utils.model_tables import Cell
+from brambling.views.utils import FinanceTable
+from brambling.tests.factories import TransactionFactory, EventFactory, \
+    PersonFactory, OrderFactory
+
+
+class FinanceTableTestCase(TestCase):
+
+    def setUp(self):
+        self.order = OrderFactory(code='TK421')
+        self.event = EventFactory()
+        self.transactions = [TransactionFactory(order=self.order)]
+
+    def test_assert_true(self):
+        assert True
+
+    def test_headers(self):
+        table = FinanceTable(self.event, self.transactions)
+        self.assertEqual(len(table.headers()), 8)
+
+    def test_plaintext_row_count(self):
+        # headers are considered part of the plaintext table
+        table = FinanceTable(self.event, self.transactions)
+        self.assertEqual(len(list(table.plaintext_rows())), 2)
+
+    def test_cell_row_count(self):
+        # cells do not include headers
+        table = FinanceTable(self.event, self.transactions)
+        self.assertEqual(len(list(table.cell_rows())), 1)
+
+    def test_transaction_created_by_blank(self):
+        name = FinanceTable.created_by_name(TransactionFactory())
+        self.assertEqual('', name)
+
+    def test_transaction_created_by_name(self):
+        creator = PersonFactory(given_name='Leia', surname='Organa')
+        created_transaction = TransactionFactory(created_by=creator)
+        name = FinanceTable.created_by_name(created_transaction)
+        self.assertEqual('Leia Organa', name)
+
+    def test_transaction_with_order_code(self):
+        code = FinanceTable.order_code(self.transactions[0])
+        self.assertEqual('TK421', code)
+
+    def test_transaction_without_order(self):
+        transaction = TransactionFactory(order=None)
+        code = FinanceTable.order_code(transaction)
+        self.assertEqual('', code)
+
+    def test_transaction_as_cell_row(self):
+        table = FinanceTable(self.event, self.transactions)
+        row = list(table.cell_rows())[0]
+        self.assertEqual(len(row), 8)
+        self.assertEqual(row[4].field, 'order')
+        self.assertEqual(row[4].value, 'TK421')
--- a/brambling/tests/functional/test_finance_table.py
+++ b/brambling/tests/functional/test_finance_table.py
@@ -1,9 +1,8 @@
 from django.test import TestCase
 
-from brambling.utils.model_tables import Cell
 from brambling.views.utils import FinanceTable
-from brambling.tests.factories import TransactionFactory, EventFactory, \
-    PersonFactory, OrderFactory
+from brambling.tests.factories import (TransactionFactory, EventFactory,
+                                       PersonFactory, OrderFactory)
 
 
 class FinanceTableTestCase(TestCase):
--- a/brambling/tests/functional/test_finance_table.py
+++ b/brambling/tests/functional/test_finance_table.py
@@ -11,46 +11,38 @@ class FinanceTableTestCase(TestCase):
         self.order = OrderFactory(code='TK421')
         self.event = EventFactory()
         self.transactions = [TransactionFactory(order=self.order)]
-
-    def test_assert_true(self):
-        assert True
+        self.table = FinanceTable(self.event, self.transactions)
 
     def test_headers(self):
-        table = FinanceTable(self.event, self.transactions)
-        self.assertEqual(len(table.headers()), 8)
+        self.assertEqual(len(self.table.headers()), 8)
 
-    def test_plaintext_row_count(self):
-        # headers are considered part of the plaintext table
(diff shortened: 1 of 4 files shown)


brambling/views/utils.py
```diff
@@ -225,3 +227,61 @@ def get_context_data(self, **kwargs):
...
+    def order_code(cls, transaction):
+        if transaction.order:
+            return transaction.order.code
+        else:
+            return ''
+
+    @classmethod
+    def created_by_name(cls, transaction):
+        if transaction.created_by:
+            return transaction.created_by.get_full_name
```
melinath: This is a method, so you'll need to call it.

melinath:
I like the idea of having a python wrapper which handles the conversion; could you go ahead and convert the template-based version to use this object as well? Also... as much as we've been bad about tests in the past, it is something that we should do moving forward. Let me know if you have any questions about how we handle / organize those.

chigby:
Also, this was very helpful in writing the tests: https://gist.github.com/NotSqrt/5f3c76cd15e40ef62d09

Otherwise, setting up the database and running migrations took over a minute each time the test suite ran.

melinath:
That does look like a helpful hack! I've been looking for something like that for a while now... That being said, I appreciate that you didn't include it in this PR. :-) I also did a manual squash of our migrations, so they *should* run quicker now...

brambling/tests/functional/test_finance_table.py
```diff
@@ -0,0 +1,57 @@
+from django.test import TestCase
+
+from brambling.utils.model_tables import Cell
+from brambling.views.utils import FinanceTable
+from brambling.tests.factories import TransactionFactory, EventFactory, \
+    PersonFactory, OrderFactory
```
melinath: In keeping with [PEP8](https://www.python.org/dev/peps/pep-0008/#maximum-line-length), we prefer parentheses to backslash. For most cases we indent the continuation line just past the paren on the start line. If there are a *very* large number of items (see urls.py) we instead put each item on a separate line.

brambling/tests/functional/test_finance_table.py
```diff
@@ -0,0 +1,57 @@
...
+
+class FinanceTableTestCase(TestCase):
+
+    def setUp(self):
+        self.order = OrderFactory(code='TK421')
+        self.event = EventFactory()
+        self.transactions = [TransactionFactory(order=self.order)]
+
+    def test_assert_true(self):
+        assert True
```
melinath: What's the purpose of this test?

brambling/views/utils.py
```diff
@@ -225,3 +228,77 @@ def get_context_data(self, **kwargs):
...
+            Cell(field='processing_fee',
+                 value=self.money(transaction.processing_fee)),
+        ]
+
+    def format_timestamp(self, transaction):
+        tz = pytz.timezone(self.event.timezone)
+        localtime = timezone.localtime(transaction.timestamp, timezone=tz)
+        return localtime.strftime("%B %d, %Y %H:%M:%S")
+
+    @classmethod
```
melinath: It would be more in keeping with Django's general behavior – admin, feeds, etc – (and the general behavior of Dancerfly) to switch these to instance methods rather than class methods.

brambling/views/utils.py
```diff
@@ -225,3 +228,77 @@ def get_context_data(self, **kwargs):
...
+
+    def cell_rows(self):
+        for transaction in self.transactions:
+            yield self.format_transaction(transaction)
+
+    def transaction_rows(self):
+        return islice(self.all_rows(), 1, None)
+
+    def transaction_as_cell_row(self, transaction):
+        return self.format_transaction(transaction)
```
melinath: We only need this *or* format_transaction.

brambling/views/utils.py
melinath: ... also more consistent internally.

brambling/views/utils.py
```diff
@@ -225,3 +228,77 @@ def get_context_data(self, **kwargs):
...
+        yield self.headers()
+        for transaction in self.transactions:
+            yield [c.value for c in self.format_transaction(transaction)]
+
+    def cell_rows(self):
+        for transaction in self.transactions:
+            yield self.format_transaction(transaction)
+
+    def transaction_rows(self):
+        return islice(self.all_rows(), 1, None)
```
melinath: `all_rows` doesn't actually exist, though I think I saw it in a previous iteration. Pretty sure this method is obviated.

brambling/views/utils.py
```diff
@@ -225,3 +228,77 @@ def get_context_data(self, **kwargs):
...
+            'Order',
+            'Amount',
+            'Dancerfly Fee',
+            'Payment Fee',
+        ]
+
+    def header_row(self):
+        return [self.headers()]
+
+    def plaintext_rows(self):
```
melinath: I think your instinct here was right. The explicit methods are probably an improvement over the magical iteration that happens in the ModelTables. We should probably port this over at some point. That being said, I wish there were a better naming convention than `plaintext_rows`... Maybe:

```python
def rows(self, include_headers=False):
    if (include_headers):
        # etc
```

melinath: Or possibly `get_rows` instead.

brambling/tests/functional/test_finance_table.py
harrislapiroff: WAHT IF ITS NOT TRUE

chigby: Hah, this was my test to make sure the test runner had found my file (which initially it didn't)! Seems worth deleting.

chigby:
```python
from cmath import pi, log
assert [0, 1, 2, 3, 4][[4, 3, 2, 1, 0][1+2+3+4-12]] == int(((2j**2)*log(1j**1j)/(2*pi)).real)
```

brambling/views/organizer.py
```diff
@@ -1009,5 +1009,26 @@ def get_context_data(self, **kwargs):
...
+
+        format_ = self.request.GET.get('format', default='html')
+
+        if format_ == 'csv':
+            context = super(FinancesView, self).get_context_data(**kwargs)
+            table = FinanceTable(self.event, context['transactions'])
+
+            pseudo_buffer = Echo()
+            writer = csv.writer(pseudo_buffer)
+            response = StreamingHttpResponse((writer.writerow([unicode(cell.value) for cell in row])
```
melinath: I think this should maybe stay unicode(cell), since that has some extra internal handling for iterables? Though we don't (and probably won't) need that for this table. But we do use it for ModelTables, and I'd like to minimize the differences between the APIs. (Alternately, convince me that we should move in this direction for ModelTables?)

brambling/views/utils.py
```diff
@@ -225,3 +228,66 @@ def get_context_data(self, **kwargs):
...
         return context
+
+
+class FinanceTable(object):
+
+    def __init__(self, event, transactions):
+        self.event = event
+        self.transactions = transactions
+
+    def headers(self):
```
melinath: I think you could potentially replace `headers()` and `header_cells()` with a `headers` class attribute. Though if the idea is to provide space for flexibility down the line, that could be good. The convention would be `get_headers()` in that case.

melinath:
A couple comments above ^ but not blocking things. I'm gonna go ahead and merge this. Congratulations!

Issue #575, opened 2015-08-20
Display Facebook URL field for events


melinath:
For some reason this isn't showing up on the event form any more... we should investigate & fix this!

melinath:
Oh hey, this was totally closed with the merge of #576!

PR #576, opened 2015-08-25
Facebook URL needs to here to be in the form. Refs #575

chigby:
I put `facebook_url` in the `EventForm`'s list of fields and, ta-da, it shows back up on the update/create event forms. 

Also, I did a little history-diving and it's not totally clear to me if this field used to be in the list and was removed, or never was there to begin with. So, I hope its being gone was just an oversight and not something intentional. That is, I'm not totally sure this addition doesn't unexpectedly change something besides the event forms. But I'm fairly sure?


melinath:
Yeah, this one is probably really this simple. Just needed someone to actually look at it and think about it for a bit to be relatively sure. :-) I'm pretty sure that the field used to be in the form, but I'm not sure when or why it was removed. Pretty sure it was an accident, though.

Issue #579, opened 2015-09-02
Add list of purchased boughtitems per attendee on attendee tables.


melinath:
This would be in place of the cart items / purchased items / refunded items fields which really aren't that useful here anyway.

melinath:
This is totally finished, not sure why it's still open. Woo!

PR #587, opened 2015-09-04
Sort orders and attendees by date

chigby:
See #579 for motivation.

So, this is not quite done yet (I think). But I'm putting the code up here because I can't figure something out with Django's template filters. Specifically the date filter. I'd sort of rather have the date formatting stuff happen in the template, but I wasn't able to get any text to show up when I applied the filter.

If it's not something we care that strongly about, then great. Though it might be worth thinking about adding some date & time specific formatting stuff to `ModelTable` that handles basic human-friendly formatting (e.g. we will probably never want to show microseconds) and maybe also time zone stuff.

### DIFF
--- a/brambling/utils/model_tables.py
+++ b/brambling/utils/model_tables.py
@@ -16,6 +16,7 @@ import six
 
 from brambling.filters import FloppyFilterSet, AttendeeFilterSet, OrderFilterSet
 from brambling.models import Attendee, Order, BoughtItem
+from brambling.utils.timezones import format_as_localtime
 
 
 __all__ = ('related_objects_list', 'ModelTable',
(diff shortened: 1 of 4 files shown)


melinath:
+1 on timezone... I think CSV can *probably* safely have "raw" data like microseconds?

brambling/utils/model_tables.py
```diff
@@ -713,3 +727,9 @@ def housing_categories(self, obj):
 
     def balance(self, obj):
         return format_money(obj.balance or 0, self.event.currency)
+
+    def completed_date(self, obj):
+        if obj.completed_date:
+            return obj.completed_date.strftime("%Y-%m-%d %H:%M")
+        else:
+            return "n/a"
```
melinath: Probably best to capitalize (N/A)

brambling/utils/model_tables.py
```diff
@@ -713,3 +727,9 @@ def housing_categories(self, obj):
 
     def balance(self, obj):
         return format_money(obj.balance or 0, self.event.currency)
+
+    def completed_date(self, obj):
+        if obj.completed_date:
+            return obj.completed_date.strftime("%Y-%m-%d %H:%M")
```
melinath: This is fine, for now, given that the housing date is handled similarly. But this does need to be wrapped in a timezone context manager, like what I think you did for the finances csv?

chigby: Right, if we're fine with not having it in the template, then I can do the same thing as with the finances table.

melinath: Yeah, we're fine not having it in the template.

PR #589, opened 2015-09-08
Switched fully-discounted transactions to use NONE instead of FAKE.

melinath:
The NONE transaction type is newer and more clearly represents what's going on.


melinath:
Just had a person be confused about this.

melinath:
this was merged in 904d120... not sure why it's not closed.

melinath:
@chigby it looks like you did a rebase, then merged, but never pushed the altered version, which confused the merge process.

chigby:
Oops, yeah, I think I misfired a git command somewhere in there. Glad it's not too hard to fix.

Issue #578, opened 2015-09-01
Better 404 and 500 pages


nmorduch:
They were cute (to me) initially but they should probably tell people to email us...

melinath:
Resolved with merge of #593.

PR #593, opened 2015-09-09
Informative error pages

chigby:
See issue #578 for original issue.

Adds a couple options, including the support email, to the 404 and 500 pages. Also includes an explicit link to the home page, in case people don't know what's going on at all. I actually feel like there could be more text here, given that it's highly possible people will encounter these pages via links from external sites (e.g. Dance organizations). If someone clicks, say, an outdated "purchase tickets" link on a totally different website, they could be very confused.

So it might make sense in the future to have more fine-grained errors like "Organization not found" or "Dance event not found" or "Dance event expired" -- maybe these already exist? -- to combat the problem.

But for now, here are some changes that at least make the error pages marginally more useful. Tonally they are somewhat bland but maybe other people have a better grasp of the exact level of cuteness we want to have here.

### DIFF
--- a/brambling/templates/404.html
+++ b/brambling/templates/404.html
@@ -1,18 +1,26 @@
 {% extends 'brambling/layouts/12.html' %}
 
+{% load staticfiles %}
+
 {% block title %}404 Error – {{ block.super }}{% endblock %}
 
 {% block main %}
-<h1>404 Not Found</h1>
-<p>No page was found at:</p>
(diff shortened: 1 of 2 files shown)


melinath:
I've made some edits to style the pages a bit more & tweaked the language a bit in a couple places. These are just suggestions that you're absolutely free to dismiss. But if you do like them, feel free to merge as far as I'm concerned. :-)

Issue #594, opened 2015-09-10
Can't edit events


melinath:
Big issue! Based on a report from a user, I think we introduced a bug whereby events can't be edited. Namely, when we check for uniqueness of event slug (#588) I think we don't exclude the current event from the qs. I haven't checked but this is my instinct.

Original report:

>Trying to update the registration information box asks me for a new slug for the event and does not allow the small change in wording.

melinath:
Resolved with merge of #595

PR #595, opened 2015-09-10
Exclude current event (if present) from slug uniqueness check. Refs #594

chigby:
This worked when I tested it. I hope it's not too un-django-ish!


melinath:
I'd suggest doing the org/event slug filter first, then excluding the pk if there is one, them running exists on that. A bit cleaner.

PR #618, opened 2015-09-13
Claim orders

### LINKED ISSUE #246: Allow users to claim their previous orders upon creating an account.

melinath:
Resolved #246. Added support for users "claiming" orders for events that they registered for anonymously. This implementation explicitly doesn't support claiming orders if the person, say, registered twice, once anonymously and once not.


melinath:
This is half of plugging the security hole around anonymous users. Once it's finalized & merged, we'll be able to do the other half: directing folks to claim orders when they verify their account, and directing folks to sign up / log in when they try to visit the link for an anonymous order that they haven't claimed.

chigby:
I've tested this out and it seems to work fine from a technical standpoint, and therefore I'm happy to merge it. But I do have some concerns.

1. I'm not totally sold on the terminology. The box says "claim an order you made anonymously." Do we really mean something like "unregistered" instead of "anonymous"? We potentially have a lot of information about the person submitting the order (email address, phone number, home address, etc.) so it doesn't feel particularly anonymous. Also I don't find that phrase being used anywhere else on the site or in the order process -- maybe "... you made before you signed up" or "... made without signing up" is more intuitive?

Also, to me, "claim" has an implication that the article in question is disputed or somehow up for grabs. As in, "if no one claims this suitcase, we're going to detonate it out back." Does "claim your order" imply that if I *don't* claim it, it's not valid, or that someone else might claim it instead of me?

That said, I'm not sure I have a better choice of words here. Ultimately the main benefit to the user is that they can see orders they made in the past more easily than digging around in their email (and possibly other things I'm unaware of?). There's not really a verb for that.

2. Which leads me to the second concern, which is that this whole process felt opaque. I wasn't sure what claiming an order would do or why I would want to do it, except a sort of uneasy urgency brought on by the word "claim" which made me feel like I ought to be doing it. Correct me if I'm wrong, but we could be doing this process automatically, and instead of that we have this claiming UI to make it more explicit and transparent what's going on. I'm not sure it's achieved that yet.

This could probably be fixed with some text explaining what's going on and what happens when I click on "Claim it!" As it is there were almost no clues.

3. And, for what it's worth, I'd also be fine with this process happening automatically on sign up.

melinath:
100% agree that "anonymous" is not a good word. I tried to avoid it in some places, but still used it in others. I'll go through and clean that up.

There are a couple scenarios that make me leery of doing this process automatically:

1. If there's a conflict between an order created without an account and an order created with an account, it would need to be resolved manually.
2. I could see someone registering for an event without logging in – say, in an incognito window on a friend's computer – then attaching the order to their account later.
3. I could see someone changing their email address (though we could do the attaching as part of the email confirmation process there.)

Maybe I'm overthinking this. But I do think that this implementation works for now (with improved copy). I'd rather merge (with improved copy) and then switch to automatic later than try to find time to implement an automatic version we'd be happy with. :-p

(I'd be interested in hearing @harrislapiroff weigh in on the automatic / not question.)

melinath:
@chigby text updates done.

harrislapiroff:
I mentioned some of this in channel, but I'm not sold on the "link" terminology. That feels overly technical to me. For the record, here is what it looks like on Humble Bundle:

screenshot 2015-09-15 18 07 59

melinath:
They use claim! #vindication

harrislapiroff:
That having been said, I think a better flow for this process might make the language less important. I realize this is more complex, but I'd be in favor of a semi-automated process where you get maybe a nice notification when you first verify your email address saying something like: "Thanks for registering! We found 2 orders associated with your email address. View orders." Or something.

harrislapiroff:
("Better" than what I'm imagining. For the record, I have not actually looked at this yet and will probably have more specific ideas once I do.)

harrislapiroff:
(Also, I didn't realize I had those Humble Bundle purchases not associated with my account...guess I should claim them. :stuck_out_tongue_closed_eyes:)

harrislapiroff:
**Notes (for this branch or others):**

- [ ] Button to resend confirmation email directs you back to homepage
- [ ] Homepage doesn't display messages

harrislapiroff:
I changed the UI. "Link orders" no longer appears in the sidebar. Instead you get this notification:

screenshot 2015-09-15 23 02 31

screenshot 2015-09-15 23 03 15

harrislapiroff:
I updated the language and UI back to claim orders, but changed the flow to make it more clear what that means. Here also are some questions/concerns for the future or now:

- [ ] 1. I did not change the code language back to "claim"
- [ ] 2. Currently the notification is not dismissable as long as there are claimable orders.
- [ ] 3. Currently this notification lives on *all* settings pages. Eventually it should probably live on an order history page. (Which would make 2 less necessary and maybe not worth fixing now.)
- [ ] 4. It might be nice if you could see that there _are_ unclaimed orders, even if you email address isn't confirmed. Just not be allowed to view or claim them. (That could theoretically be considered a security hole, since you could determine _if_ someone had orders, but that seems like a pretty small potential information leak.)

harrislapiroff:
Bouncing this back to either @chigby or @melinath for review or further work.

melinath:
I'm gonna change the code language back to 'claim', then do a final functionality pass, then merge if all looks good.

melinath:
This is weird... should the number up top be a sum of both lengths?

screen shot 2015-09-16 at 9 52 38 am

harrislapiroff:
ohdang, probably

harrislapiroff:
actually, wait. Probably the first message just shouldn't be displayed.

melinath:
but if you have claimable *and* unclaimable orders, that first sentence will still be kinda weird.

harrislapiroff:
Yeah. Maybe both changes. The first message doesn't really seem critical if there are only unclaimable items, since there's no action that can possibly be taken on this page anymore, haha.

melinath:
Okay, I think that should do it. Bouncing back to @harrislapiroff or @chigby for review.

chigby:
I ran through this again and it is looking great. Much clearer about what is going on. I'm fine to merge it if @harrislapiroff thinks so too.

harrislapiroff:
:+1:

PR #620, opened 2015-09-14
Implemented a view to remove attendees from an order.

melinath:
All you have to do is delete the attendee and the BoughtItems become unassigned automatically. :-)


chigby:
Everything seems to be good, here!

PR #621, opened 2015-09-14
Implemented an item delete view. Resolved #533.

### LINKED ISSUE #533: Item Deletion
There's currently no way to delete an item, but there should be.

melinath:
I know that we "shouldn't" use get for deletion views, but I think this is a case of practicality beats purity. Getting a browser to *actually* issue a POST or DELETE for a deletion is a PITA. :-p This is a simple implementation that will let organizers delete items whenever they want.


harrislapiroff:
Wait, all you have to do to fire a POST is make a form with a button! I would somewhat strongly advocate for that.

harrislapiroff:
For the record that is also how Django's generic deleteviews do it and it protects the user in a host of ways including CSRF. (I can't maliciously send a user a link that will delete their data.)

melinath:
The button is inside a form already! I can't create a form in a form.

harrislapiroff:
Then move it after the form. Upon consideration, I would actually call this a pretty severe security hole.

melinath:
If it weren't then yeah I would've done that. And yeah, of course I could work around this by, like, dynamically creating a form or changing the action of the form and deleting all the fields, but that level of weirdness just seems bizarre.

melinath:
If it's after the form it doesn't flow correctly with the save button. Unless I hook that up with JS to submit the form properly.

melinath:
Which would be less weird, I suppose.

harrislapiroff:
Any solution is better than letting me, a malicious dance event competitor, send you a link which triggers a permanent delete of your valuable attendee data.

harrislapiroff:
(Or letting a browser think it's OK to hit that URL over and over again, though in this case that's less critical since that URL will only work once anyway.)

melinath:
I'll fix this in a bit.

harrislapiroff:
I'm pretty sure I have javascript in the repo already for triggering submission of another form based on clicking a button elsewhere, but I can't remember where.

melinath:
It's not that hard. I can write that JS.

melinath:
@harrislapiroff fixed

chigby:
I got a little squeamish about having a delete button without a confirmation box, so I added one. Hope my little explanation is correct, but if anyone wants to change the wording or something, they should.

melinath:
I almost put in a confirmation box but then I was too lazy :-p

harrislapiroff:
:+1: to confirmation modals for deletion

plus it solves the nested forms issue :stuck_out_tongue_closed_eyes:

melinath:
I already solved the nested forms issue! :-p

harrislapiroff:
Now your solution is excessive!

melinath:
UGH.

melinath:
image

melinath:
fed-up-with-this-world-o

melinath:
done

melinath:
screen shot 2015-09-16 at 1 59 25 pm

PR #625, opened 2015-09-16
Dashboard refinements:

harrislapiroff:
* Moved title and hint into main column instead of above both columns
* Added a block for messages


harrislapiroff:
Before:
screen shot 2015-09-15 at 23 37 22

After:

screen shot 2015-09-15 at 23 37 35

PR #626, opened 2015-09-16
Some form UI updates:

harrislapiroff:
* Marking optional fields with "(Optional)" instead of required
  fields with asterisks. See: http://www.lukew.com/ff/entry.asp?725
* Chosen multiple select placeholder
  "Select Some Options" -> "Select all that apply"
* Chosen single select placeholder
  "Select an Option" -> "Select one"

### DIFF
--- a/brambling/forms/orders.py
+++ b/brambling/forms/orders.py
@@ -84,7 +84,7 @@ class AttendeeBasicDataForm(CustomDataForm):
                 (Attendee.HOME, 'Staying at own home'),
             )
             self.fields['housing_status'].initial = ''
-            self.fields['phone'].help_text = 'Required if requesting housing'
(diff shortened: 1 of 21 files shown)


harrislapiroff:
This does reveal some weird things about our form handling that should maybe be rectified:

screenshot 2015-09-16 11 18 58

1. The "(Optional)" marker with the phone field is a little awkward with the help text. (The help text is only rendered when housing is enabled.) I'm not sure what exactly is a better solution than this.
2. Items are optional?

melinath:
I think this is an improvement and would be happy to merge as-is. What would you think of switching the phone # help text to "Optional unless requesting housing?"

harrislapiroff:
I'm also happy if you want to merge this as is. The help text change you're suggesting also seems reasonable to me.

Issue #558, opened 2015-07-22
CSV Issues


harrislapiroff:
@spilzer mentioned to me the other day that there might be some issues with Excel reading Dancerfly-produced CSVs? Or maybe it was with Google Sheets reading Dancerfly-produced CSVs? I can't remember. Either way, this should be investigated.

melinath:
Just got this from a user:

>When attempting to download csv I get this message "file format & export csv don't match. The file could be corrupted or unsafe...."
>And when I click on it anyway
>"It is a SYLK file but cannot load it"

harrislapiroff:
That tracks with what @spilzer told me. Sounds urgentish. Will look into soon.

melinath:
The user looked into it some more and found the following: https://support.microsoft.com/en-us/kb/323626

And was able to open the file following those instructions... which were for Office 2003. Do we want to support versions that old? Maybe we should just instruct people having issues to use libreoffice or google drive?

melinath:
Or we can just be fine with the xlsx alternative being available.

harrislapiroff:
I want to look into if we can either write XLS files or if the excel dialect of the python CSV writer fixes this.

harrislapiroff:
We don't currently provide xls do we?

melinath:
No; ticket #572 is currently tracking that issue.

harrislapiroff:
Ah ha

melinath:
Resolved with merge of #628

PR #628, opened 2015-09-17
Fix file-format bug from Excel 2003. Refs #558

chigby:
Apparently Excel doesn't like to open files that begin with "ID", so
let's change it.


melinath:
I don't suppose you were able to test this?

PR #629, opened 2015-09-17
Made finances table use Rows so that we can still access the transaction object.

melinath:
Turns out the finance CSV changes broke the links to remote charge URLs! This should fix the issue.


chigby:
I see the problem.. I guess it didn't show up when I tested it because none of my fake transactions have remote transaction URLs!

melinath:
Yeah... I don't have that excuse for not noticing it on review... :-p

melinath:
So it goes. ¯\_(ツ)_/¯

Issue #572, opened 2015-08-17
Offer .xls files


harrislapiroff:
Yeah, .csv files should be sufficient but people like to get things in familiar formats. (And this would ensure proper import, which people have maybe been having trouble with? #558) We could use https://pypi.python.org/pypi/xlwt

chigby:
Here's the thing: are we sure we need this?
 
 * Right now, our data is just columns and rows. We are not taking advantage of any Excel-specific features or formatting. Changing the format would be entirely cosmetic.
 * Notwithstanding a bug in a 10 year old piece of software (#558), CSV files are supported by all versions of Excel, as well as Open Office, Numbers, Google Drive, Gnumeric, awk, and whatever else. XLS files are a less compatible version of the same thing.
 * Is XLS more familiar to people than XLSX, which has been the format Excel has used since 2007. Do we want to offer both? Could lead to even more confusion.
 * If people are interesting in doing interesting things to our data with Excel, let's find out what those things are. Maybe we could help? Offer derived numbers natively, provide an integration to services that they're piping the numbers into, offering Excel-specific spreadsheets with formulas and stuff that are useful? Add some value to what they're doing, you know!

melinath:
If we were to do this, it would be .xlsx for maximum familiarity. Personally I'd be fine with leaving it as just CSVs for now.

harrislapiroff:
I think it's pretty important to provide people formats that they're familiar with, regardless of functionality, for both convenience and comfort's sake. A person's computer is not guaranteed to automatically know what to open a CSV file in, but if it has Excel installed, it will know how to open an XLS. And Excel, unlike Numbers and the rest, is an "industry standard" :stuck_out_tongue_winking_eye:. Our users are not necessary tech-savvy and the word CSV throws many people.

XLSX and not XLS would be sufficient from my perspective.

harrislapiroff:
Ftr here are the formats Google Sheets provides. I don't think we need to go this far, but you can see they also provide multiple formats that have essentially the same functionality (esp. TSV and CSV)

![screenshot 2015-09-17 17 19 14](https://cloud.githubusercontent.com/assets/296631/9946286/4ddb427c-5d60-11e5-9303-4cddbe9366d1.png)

harrislapiroff:
https://pypi.python.org/pypi/XlsxWriter looks nice and pretty easy

chigby:
[...]

PR #631, opened 2015-09-18
Add Excel file export. Refs #572

chigby:
Add a button to export the currently exportable tables into .xlsx format.

The UI is about as simple as possible and could possibly be improved, but I am not really familiar enough with the buttons and controls you have going so it might be inconsistent. Having one button for csv and one for excel could pretty easily be replaced with a dropdown or popup or something else to make it fit in better with what you have or just generally be more user-friendly.

### DIFF
--- a/brambling/static/brambling/sass/modules/_menus.sass
+++ b/brambling/static/brambling/sass/modules/_menus.sass
@@ -1,2 +1,20 @@
 .dropdown-menu > li > a
 	padding: 3px #{$line-height-computed/2}
+
+
+//https://github.com/twbs/bootstrap-sass/blob/master/assets/stylesheets/bootstrap/_dropdowns.scss#L65
+.dropdown-menu > li > .btn-link
+	display: block
+	padding: 3px 20px
+	clear: both
+	font-weight: normal
+	line-height: $line-height-base
+	color: $dropdown-link-color
+	white-space: nowrap
+	width: 100%
+	text-align: left
+
+	&:hover, &:focus
+		text-decoration: none
+		color: $dropdown-link-hover-color
+		background-color: $dropdown-link-hover-bg
--- a/brambling/templates/brambling/event/organizer/__table.html
+++ b/brambling/templates/brambling/event/organizer/__table.html
@@ -81,6 +81,10 @@
                         <i class="fa fa-download"></i>
                         <span class='hidden-sm hidden-md'>Download CSV</span>
                     </button>
+                    <button type="submit" class="btn btn-sm btn-default tipped" name="format" value="xlsx" title="Download Excel" data-container="body">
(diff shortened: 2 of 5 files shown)


harrislapiroff:
Either the buttons as is or a dropdown would be fine with me. Dropdown might be trickier since the elements inside it have to be `<buttons>` and I'm not 100% sure how Bootstrap will handle that. Could also be done as a popover the way I did the saved reports dropdown. But I'm also fine with it as is!

melinath:
Could just be a dropdown like toggl.

screen shot 2015-09-18 at 10 35 12 am

melinath:
screen shot 2015-09-18 at 10 36 32 am

Or Google Docs. No need for buttons.

harrislapiroff:
They need to be buttons because they're submitting a form, is why. :stuck_out_tongue_closed_eyes:

harrislapiroff:
(Though the whole system that's going on with the form that navbar is a UX mess and should be restructured eventually. Maybe when we make it async though.)

melinath:
Oh right. :-p

FWIW I am also getting weird flow on the order table.

screen shot 2015-09-18 at 10 39 32 am

melinath:
I'ma give it a shot.e

nmorduch:
looks good but when i tried to use it i got 

>Error response

>Error code 414.

>Message: Request-URI Too Long.

>Error code explanation: 414 = URI is too long..

nmorduch:
But if it works (eg just wasn't working for me) :+1:

melinath:
@nmorduch was that on the attendee view? Slash which browser?

nmorduch:
orders, FF 40

melinath:
So, I suspect that's probably due to a long querystring on the order view – from all the columns etc. – which would be a separate ticket. Could you try on the finances view?

harrislapiroff:
Time to switch to POST? :|

melinath:
so it would seem. but in a separate ticket.

melinath:
http://thedailywtf.com/articles/HTTP-414-Way-Too-Fing-Long

attn @nmorduch especially :-) This is actually the first time I've ever seen a 414 error! Exciting times! I can't reproduce it myself, but I'm confident it has nothing to do with the excel download portion of this ticket. I'll open a new ticket for that issue.

harrislapiroff:
OOC what is responsible for serving the 414? nginx? Django? Not the browser, I assume? Is this a server-side setting we can just change. Also, how long was the URI???

melinath:
It's a server-side thing. http://stackoverflow.com/questions/1067334/how-to-set-the-allowed-url-length-for-a-nginx-request-error-code-414-uri-too

I had one that was 1043 characters long working fine. I'd have to work pretty hard to get over 2900 characters... I think we should be fine for now. But someone will break this someday.

PR #634, opened 2015-09-19
Improved accuracy of attendee table balances & item lists while reducing query counts.

melinath:
Attendee table was displaying the total order balance - but something in the query was fucking shit up and showing balances two, three times as large as the actual order balance. Plus, that information doesn't make any sense given the table's focus on the attendee object itself. So I removed that column altogether and instead included how much had been paid for the attendee's items.

Future improvement along these lines: add boolean column for whether there are unconfirmed payments (or with the amount of unconfirmed payments). My only concern is that detecting whether there are unconfirmed payments / how many there are might be a PITA, since transactions are order level. But it can probably be done... as a separate PR.

### DIFF
--- a/brambling/utils/model_tables.py
+++ b/brambling/utils/model_tables.py
@@ -6,7 +6,7 @@ from django.contrib.admin.utils import (lookup_field, lookup_needs_distinct,
                                         label_for_field)
 from django.contrib.admin.views.main import EMPTY_CHANGELIST_VALUE
 from django.core.exceptions import ImproperlyConfigured
-from django.db.models import Q, Sum, Min
+from django.db.models import Q, Sum, Min, Prefetch
 from django.forms.forms import pretty_name
 from django.utils.datastructures import SortedDict
 from django.utils.text import capfirst
@@ -395,7 +395,7 @@ class AttendeeTable(CustomDataTable):
         ('Identification',
          ('pk', 'get_full_name', 'given_name', 'surname', 'middle_name')),
         ('Status',
-         ('items', 'purchase_date')),
+         ('items', 'purchase_date', 'paid')),
         ('Contact',
          ('email', 'phone')),
         ('Housing',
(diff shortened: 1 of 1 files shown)


brambling/utils/model_tables.py
```diff
@@ -482,8 +485,13 @@ def order_placed_by(self, obj):
...
 
-    def order_balance(self, obj):
-        return format_money(obj.order_balance or 0, self.event.currency)
+    def paid(self, obj):
+        paid = 0
+        for item in obj.items:
+            paid += item.price
+            for discount in item.discounts.all():
+                paid -= discount.savings()
+        return format_money(paid, self.event.currency)
```
chigby: I feel like this could be a little more compact, e.g.

```python
paid = sum(i.price for i in obj.items)
paid -= sum(d.savings() for d in obj.items.values_list('discounts', flat=True) if d)
```
Which is possibly a bad/nonstandard way to query up those discounts but it's what I came up quickly.. maybe you know of a better way?

I mean, we are adding up numbers. Surely there's more efficient way than nested for loops?

chigby:
This seems fine to merge, even without revising the summation function. If it's important I say go ahead with it.

brambling/utils/model_tables.py
melinath: So the trick here is that `item.discounts.all()` is [prefetched](https://github.com/littleweaver/django-brambling/pull/634/files#diff-64a8f85cd1df4eacbf6fca6b934ed4bfR455), so it's only one db query for the whole table. I don't *think* there's really a better way to iterate over them... but I could be missing something.

From a Django perspective, `obj.items.values_list('discounts', flat=True)` doesn't work because discounts itself goes to multiple related objects. You could try annotating the items queryset with a sum of the discount amounts, then doing values_list on *that* – I think that would work. But at that point you'd be doing one query per row, rather than one per table, which would probably be less efficient.

Separately, it would be a tricky thing to calculate how much savings a discount would give someone from a query, since the amount can either be a dollar amount or a percentage.

Hmm. Actually, if we cached the real savings amount in a field instead of re-calculating it every time, we *could* annotate the items qs with total savings, probably. But the efficiency should already be pretty good as is.

Issue #592, opened 2015-09-08
Better security for users without accounts


melinath:
Short order codes can be nice for some things, but it's not great for security for anonymous users. Really we should have 20 char codes. That would also necessitate a shorter identifier for the order, though. I'd suggest a per-event order number - essentially an index of where the order was created.

harrislapiroff:
I don't think this is actually going to improve security _that_ much. It will definitely make brute-forcing harder, but I think brute-forcing is as relevant to us as these potential attacks:

* intercepting a URL with a MiTM attack (This seems like the least likely, since we have HTTPS on, but...)
* discovering a URL in browser history. (Say someone orders, or even just _looks at_ a Dancerfly ticket on a public computer.)
* acquiring a URL from an email someone forwarded to you. (Say I forward my receipt to my rideshare group to confirm that I'm registered and someone, on a whim, clicks the link to view it on site.)

These all seem like more likely attacks to me and they won't be fixed by longer codes. I also would advise against consecutive numbered order identifiers. Using random codes means that if I am checking in with an order code I'm less likely to accidentally copy down a wrong code (say, off by one digit) and have it accepted as valid or be able to maliciously give the code of the another person. (Sure, yeah, my order number is "00043".)

harrislapiroff:
Or if I'm an organizer giving another organizer a list of order codes to be refunded, or marking a check with an order code on it is paid, etc.

melinath:
Just to have these copied over, these are security improvements along these lines that @harrislapiroff had suggested in another thread:

>1. Have the receipt be the only confirmation the user has. Don't give them a permanent URL.
>2. Automatically create an account for even anonymous users. The permanent link is now just a temporary URL. The first time they try to view/edit their information on site, force them to create a password and retire the temporary URL. (IIRC this is what eventbrite does.)
>3. Don't give users a permanent URL, but allow them to request a one-time-use temporary auth URL sent to their email address anytime, which expires after 24 hours. (This would manage all tickets associated with that email address.)

melinath:
Resolving this by first implementing #246 (claiming anonymous orders from confirmed accounts) and then requiring login / creation if someone visits the URL.

chigby:
Just a thought: would it make sense to require the order code _and_ their email address to see the order details? This is how Expedia does it.
[...]

PR #635, opened 2015-09-21
Accountless security

chigby:
See #592.

Not addressed is adding a note in the registration email that there are unclaimed orders.


brambling/templates/brambling/email_confirm.html
```diff
@@ -1,20 +1,36 @@
...
 
 {% block main %}
-	{% if valid_token %}
+	{% if valid_token and not claimable_orders %}
 	    <div class="alert alert-success">
 	        Your email address has been confirmed! You will be redirected to <a href="{% url 'brambling_dashboard' %}">your dashboard</a> in a few seconds.
 	    </div>
+	{% elif valid_token and claimable_orders %}
+		<div class="alert alert-success">
+	        Your email address has been confirmed! To use your account,<a href="{% url 'brambling_dashboard' %}">continue to your dashboard</a>.
```
melinath: Missing a space after account

brambling/views/orders.py
```diff
@@ -247,6 +247,22 @@ def dispatch(self, request, *args, **kwargs):
...
+        if 'code' in kwargs:
+            # If an anonymous user is trying to access an anonymous
+            # order without a session, redirect them to the login
+            # page.
+            matching_order = Order.objects.filter(event=self.event,
+                                                  person__isnull=True,
+                                                  code=kwargs['code']).exists()
+            if (matching_order and not request.user.is_authenticated() and
+                not Order.objects._get_session_code(request, self.event)):
+                url = reverse('login')
```
melinath: I think I might prefer in this case to display a helpful message, as well as either a login or signup form depending on whether the account already exists. (Rather than just redirecting to login.) At a bare minimum we should add a message (using `django.contrib.messages`) explaining wtf is going on.

melinath: Actually, we should probably display a helpful message any time that someone accesses an order they shouldn't have access to, and only redirect to their own order summary if this is actually their order.

melinath: This could even be a separate view, since it has a separate url pattern.

melinath: I would prefer to have this be a separate view, since it would make the code paths clearer / cleaner.

brambling/templates/brambling/mail/order_receipt/body_plaintext.txt
```diff
@@ -24,7 +24,7 @@ Total: {{ total_payments|format_money:event.currency }}
 
 Order #{{ order.code }}{# Transaction date #}
 
-{% if order.person %}{% url 'brambling_event_order_summary' event_slug=event.slug organization_slug=event.organization.slug as url %}{% else %}{% url 'brambling_event_order_summary' event_slug=event.slug organization_slug=event.organization.slug code=order.code as url%}{% endif %}
+{% if order.person %}{% url 'brambling_event_order_summary' event_slug=event.slug organization_slug=event.organization.slug as url %}{% else %}{% url 'brambling_event_order_summary' event_slug=event.slug organization_slug=event.organization.slug as url%}{% endif %}
```
melinath: this one (and the related ones in emails) actually need to direct to the newly created view, so that anonymous users end up on the page directing them to register (rather than accidentally creating a new order).

brambling/views/orders.py
```diff
@@ -306,11 +293,42 @@ def get_reverse_kwargs(self):
...
+            order = Order.objects.get(event=self.event, code=kwargs['code'])
+        except Order.DoesNotExist:
+            raise Http404
+
+        session_code = Order.objects._get_session_code(request, self.event)
+        order_in_session = kwargs['code'] == session_code
+        if (order and not order.person and not request.user.is_authenticated()
+            and not order_in_session):
+            url = reverse('brambling_signup')
+            messages.add_message(request, messages.ERROR, "Please sign up with an account to view this order.")
```
melinath: Since we're redirecting, might be good to add the order code into this message & below. Something like: "Order abcdef was created without an account. Create an account with the same email address as the order to view it."

melinath:
I feel like maaaybe this should be reviewed again, but also it does need to be merged, and I didn't change that much? Feel free to scold me in the morning if this was the wrong call.

PR #644, opened 2015-09-24
Custom housing forms

melinath:
And lo, custom housing & hosting forms!


melinath:
This was briefly smoke-tested by Cameron. I'm going to ask the event to test it as well.

Issue #287, opened 2014-12-21
Better interface for marking checks as paid.


harrislapiroff:
The current button isn't obviously a button, but perhaps more importantly, there should be probably be a box where you can enter an order number and mark it as paid, rather than having to pull up that order's detail view from the table.

melinath:
I don't think it would be very user-friendly to have a text field that you put the order's code in. Maybe as an admin action-like thing?

harrislapiroff:
@melinath Doesn't an organizer already have to put the order code in to search for it?

harrislapiroff:
Actually, wait, there's no way for an organizer to find the correct order right now?

harrislapiroff:
I think a field for searching by order code is going to be a necessary step to find the correct order to apply a payment to for any event that has more than, say, 50 orders. I can't think of any more user-friendly way to locate the correct order, though I'm open to ideas.

Creating a form that accepted an order code and marked an order as paid would just reduce the task of accepting a check to a single step instead of searching for the correct order, viewing it's details, and marking it as paid.

nmorduch:
What I would really like to see is something like a Chosen select in which each option of the select has the order number, associated email address, and attendee names. Then you can type in any of those and it will come up. This would be especially nice if you can format the options to display in multiple rows (thinking of the Facebook search bar).

Alsoooo it would be really great if selecting them pulled up their whole order info, ajax style.

Or. It could be a multiselect and and you can select all the orders that you're marking as paid right now, all in one box!

harrislapiroff:
Yeah, an order live filter would be badass. I guess I'm wondering if this is a common enough task, though, that there should be a form (for events that have check payment enabled) on the event dashboard for quick payment entry, basically.

harrislapiroff:
That quick payment entry form could also have an ajax preview of the order though, to confirm that you had typed in a valid and correct order code. That'd be pretty sweet.

nmorduch:
So Naomi had fun with box drawing and then they look sad here. Anyway, my thought is to have it in a little tabbed thing when checks are enabled because you want both *find* and *pay* options but you don't want to be tabbing through a lot of buttons; you just want to choose what mode you're in, then do a bunch of them.

```
┏━━━━━━┓───────────┐
┃ Find ┃ Mark Paid │
┃      ┗━━━━━━━━━━━┪
┃ ╔══════════╗ ┌─┐ ┃
┃ ║Find order║ │→│ ┃
[...]

PR #654, opened 2015-10-04
Order quickfind

harrislapiroff:


harrislapiroff:
I think this is mostly ready. I should mark up the component with ARIA tags, but I don't know how yet...

harrislapiroff:
Nope, I also need to do the permission thing.

melinath:
This is super fancy - love it! Minor thought, shouldn't the loading spinner go at the top? And while we're in ractive, it could show/hide with slide?

melinath:
Interestingly, this gives me access to orders that won't show up on the table normally - i.e. orders that haven't yet checked out. I don't think that should be a problem, since they're (sort of) ephemeral and also people will generally use this when they already know what they're looking for.

Is the idea that this resolves the check resolution marking issue?

brambling/api/v1/views.py
```diff
@@ -155,6 +157,37 @@ def create(self, request, *args, **kwargs):
...
+    serializer_class = OrderSerializer
+    filter_backends = (filters.SearchFilter,)
+    permission_classes = [OrderSearchPermission]
+    search_fields = ("code", "person__given_name", "person__middle_name",
+        "person__surname", "attendees__given_name",
+        "attendees__middle_name", "attendees__surname")
+
+    def get_event(self):
+        event_id = self.request.query_params.get('event', None)
+        return Event.objects.get(pk=event_id)
```
melinath: This will throw an error if event_id is None. Better to wrap in an if and return None otherwise. Or `return None if event_id is None else Event.etc` if you want a one-liner.

melinath: But I'm not super concerned... the chances of someone accessing this improperly seem low. But also it would be an easy fix.

brambling/static/brambling/brambling.utils.js
```diff
@@ -0,0 +1,30 @@
+// Some simple Javascript utilities to be used wherever:
+
+;(function () {
+	"use strict";
+
+	// Filters to be used in Ractive mustaches:
+	brambling.filters = {
+		// Wraps a bit of text in an element:
+		highlight: function (needle, haystack, open_tag, close_tag) {
```
melinath: Needs to handle the case where needle or haystack is undefined (for whatever reason). Just an `if (!needle || !haystack) return undefined` should do it. Otherwise you get warnings clogging up the console.

melinath:
Super shiny! I'd be down with merging after addressing the minor things above.

harrislapiroff:
Thanks! This doesn't fully resolve #287, because iirc the button for marking something as paid is still kinda hard to find/identify, but it's a first step/one of the ideas that came out of #287.

I was trying to figure out where the tables filtered out those orders in progress, but I couldn't find it. How does that happen? I'd like to try to make them consistent.

The loading indicator at the top of the list _might_ make sense, but I also don't want to push down the other items to make room for it (especially if they're going to end up being the same as in the case where the search goes from "Harr" to "Harri"). I'll think on it.

I'm concerned that transitions will make this _feel_ slower (and also require some handling for the case where items are the same after a search query) but I'll think on that as well.

I'll work more on this soon.

melinath:
So right now, the loading indicator is pretty much off the screen in some cases, which is weird. Plus it just feels very distant from the typing that causes it. I only think that the loading indicator should transition in, to be clear. Also, you can tweak the speed of the transition to make it feel snappier...

harrislapiroff:
Ohhh you're right, that doesn't make sense. I was testing with a low volume of orders. :stuck_out_tongue_closed_eyes:

melinath:
The filtering for the order table is just: does it have any transactions? Maybe we should be checking for bought items instead, but... not sure.

https://github.com/littleweaver/django-brambling/blob/master/brambling/views/organizer.py#L856

brambling/static/brambling/brambling.utils.js
harrislapiroff: I now have it returning undefined if the haystack isn't there, but having it return the original haystack if the needle isn't there.

harrislapiroff:
Okay, I did add a slide transition to the results. Ractive seems to be good about not rerunning the transition if the result count stays the same. This should fix all those issues. Take another look @melinath?

melinath:
I'm pretty ambivalent about the slide transition on the items in the list, but if you like it :-) The loading indicator on the field is sweet! I'll go ahead and merge.

PR #662, opened 2015-10-10
Added saved attendees

### LINKED ISSUE #233: Save multiple attendees per account rather than associating account with one attendee
Right now, accounts have names attached and save one attendee profile, but if, for example, you generally sign up as a pair, it doesn't make so much sense for only one of you to be the primary person on the account, and never save the other person's information.

#### Proposal:
Remove name from user information. Just use email/password for both login and signup.
Save a collection of attendee profiles, with the ability to remove or edit them.
In attendee/pass assignment, you can use existing attendees or create new ones.

melinath:
Resolved #233. Added a SavedAttendee model; users can save more than one attendee per account and use them to choose default settings for their attendee forms. Helpful for folks who regularly register more than one person.

Faults with this PR: I did also remove the "save as defaults" functionality, so it's even less obvious how to set up your first saved attendee. I wasn't sure how to best do the UI/UX... :-/


melinath:
Oh, probably best to copy your db before testing this branch. It doesn't have 100% reversible migrations.

harrislapiroff:
Looking at this now.

harrislapiroff:
This UI needs some love.

screenshot 2015-10-24 13 13 48

harrislapiroff:
- [ ] Columns need to be wider for attendee list view
- [ ] Probably should handle the case for very long names
- [ ] Probably should display some details about the attendee
- [ ] Have some explanation of what the point of entering an attendee into your settings is
- [ ] Change icon of attendees nav item

harrislapiroff:
Attendee selection view (referenced above): 

screenshot 2015-10-24 13 30 04

harrislapiroff:
Is there any reason not to add a checkbox back in that says something like, "Save details" that would auto-create a saved attendee? (It would probably also want a tooltip like "Save these details to be prefilled when registering for other events on Dancerfly" or something.)

harrislapiroff:
I guess there would have to be some tricky logic involved there, since if you prefilled from a saved attendee and then checked that box you'd just keep creating duplicates. Which I could see a lot of people accidentally doing.

harrislapiroff:
Also, I think I like the phrase "Prefill details" for describing the behavior of using a saved attendee. I might use that in the UI.

harrislapiroff:
We might as well use a name here in place of "New Attendee" if it's prefilled from a saved attendee.

screenshot 2015-10-24 13 37 36

melinath:
I would like to have a "save these details for later" checkbox that would either create one or update the current one, but I wasn't sure how to do it best and didn't want to make the pr more complicated than it already is.

PR #663, opened 2015-10-10
Set up disabling of buttons on payment form submission to prevent duplicate payments.

### LINKED ISSUE #658: Duplicate charges
A few people have gotten double-charged for their orders - probably from double-clicking the payment button. We should very much fix that. Probably just by doing a 'disable this button after click' thing.

melinath:
Resolved #658. As per discussion there, this should be enough, since all we need to prevent is accidents rather than malicious intent.

### DIFF
--- a/brambling/templates/brambling/event/order/summary.html
+++ b/brambling/templates/brambling/event/order/summary.html
@@ -25,6 +25,10 @@
 				set_stripe_key();
 			});
 			$(set_stripe_key);
+
+			$('form').on('submit', function () {
+				$(this).find('button').prop('disabled', true);
+			});
 		</script>
 	{% endif %}
 	<script>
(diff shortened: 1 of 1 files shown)


chigby:
Does the underlying double-click issue affect check orders too? All of my order testing has been done with check orders because I'm not sure how to set up a credit card account for testing stripe and dwolla transactions. Is that tricky?

I can at least confirm that these changes don't break ordering by check.

melinath:
It affects check orders too, yes. Can you also confirm that double-clicking doesn't create duplicates vs. master?

brambling/templates/brambling/event/order/summary.html
```diff
@@ -25,6 +25,10 @@
 				set_stripe_key();
 			});
 			$(set_stripe_key);
+
+			$('form').on('submit', function () {
+				$(this).find('button').prop('disabled', true);
+			});
```
chigby: This should be outside the `{% if event.stripe_connected %}` block.

melinath: Hooooly shit. And this is why review. Thanks!

Issue #659, opened 2015-10-09
Dwolla facilitation fees being charged?


melinath:
Ran into an error where dwolla `charge['Fees']` was `None`... not sure why. There should've been a facilitation fee. It concerns me. (Side note, this was apparently our first actual dwolla charge! Huh!)

melinath:
It definitely looks like we're not pulling in facilitator fees on our recent live transactions. But our integration tests work fine... so I'm not sure what's going on yet.

harrislapiroff:
Is this something to ask dwolla about? Do the API calls look right?

melinath:
I mean, the integration tests work fine, and they use the same api calls. So yeah, we should probably ask dwolla support.

PR #664, opened 2015-10-11
Dwolla updates

melinath:
Was hoping to test #659, but had to make some [updates to our dwolla handling](https://discuss.dwolla.com/t/action-required-oauth-redirect-uri-must-be-configured-in-application-settings/1605) first. Those should be done with this PR.


melinath:
Okay, that makes the tests runnable. We'll only have to generate new tokens if/when the api changes.

chigby:
I feel like something weird is happening with my setup, because the test fails with: 

```
FAIL: test_dwolla_charge__user (brambling.tests.integration.test_dwolla.DwollaChargeTestCase)
[...]
AssertionError: u'Order 000001 for Test event' != 'Order 000041 for Test event'
```

chigby:
Or should I be recording my own vcr cassette?

melinath:
@chigby No, you should be able to use the same cassette. My guess is that you're running the whole test suite, whereas I only ran the one test, so the order code is different... though that shouldn't be stored anywhere... hmm.

melinath:
Oh, it's because the value on the left is charge['notes'].

melinath:
I'm gonna switch to a deterministic order code.

Issue #665, opened 2015-10-12
Double clicking add-to-cart will add items twice


chigby:
If you double click on an item in the shop, it will add two of them to your cart. I think this is generally not expected and, if not noticed before checkout, could lead to incorrect purchases.

Also, does not interact well with items of limited quantity:

![screen shot 2015-10-12 at 14 28 39](https://cloud.githubusercontent.com/assets/561931/10435457/1afe5ece-70ee-11e5-9b22-c8748ab6dea4.png)

melinath:
Oh hey this was resolved  by the merge of #666.

PR #666, opened 2015-10-12
Prevent links from double-posting when double-clicked

chigby:
See #665 for motivation.

This change will stop post requests from firing twice when double-clicked. I am not sure to what extent the site is using the convention of `<a target="...">` to trigger a post, so I hope this doesn't break anything I didn't think of. Ideally, though, this will fix double-clicking in all those places.


melinath:
This doesn't quite work as-is. When the script runs the first time, it looks for all `a[target]` elements and attaches a listener to them. Removing the target attribute later doesn't remove the handler.

One alternative would be to move the click event up to the body - i.e.:

```javascript
$('body').on('click', 'a[target]', function() {});
```

(You can test by commenting out the post and logging each click event.)

chigby:
Good catch! In my manual testing it did appear to be working, but maybe my double-click finger is rusty. 

One other solution I though of (though I think yours is simpler) would be to add a "clicked" marker to the element that was clicked, then only send the post if the marker isn't there.

harrislapiroff:
What's the status of this branch? Is it ready for review or is it awaiting a solution?

melinath:
It's ready for review.

melinath:
@harrislapiroff just to clarify, this branch works, I just wasn't sure if there might be concerns with the method used.

harrislapiroff:
It's not ideal, I think, but the overloading of `target` to contain URLs is the not ideal part and should be fixed elsewhere I think. Reviewing now.

harrislapiroff:
Actually I have a more elegant way of doing this I think.

harrislapiroff:
Eh, I changed my mind.

harrislapiroff:
I was going to use jQuery's `.one` method, but because we need to `preventDefault` on every click, that doesn't actually work.

Issue #669, opened 2015-10-20
Background on banners should be white


harrislapiroff:
This rendering of a transparent PNG banner is not ideal. (Also not consistent with the dashboard appearance.)

![screenshot 2015-10-20 12 24 00](https://cloud.githubusercontent.com/assets/296631/10613939/92b73156-7725-11e5-8eb9-d69c62b7a410.png)

PR #671, opened 2015-10-21
Ensure a white background for banner images.

chigby:
See #669 

Mostly this is for better compatibility with transparent PNG images.


melinath:
@chigby side note, we try to stick to past tense for commit messages. I.e. "Resolved" rather than "resolves"

Issue #473, opened 2015-04-25
Improve attendee / order table cell widths


melinath:
Right now, the table cells use the default behavior of shrinking as much as possible, which leads to really weird text flow. It would be great if we could make this behave better... not sure whether that's possible.

harrislapiroff:
In my ideal world these tables will be more sophisticated, like a spreadsheet. In the meantime, we could use CSS to prevent text wrapping and even provide a button to toggle that on and off.

melinath:
I don't think we want to *prevent* text wrapping, more like have a minimum width on all the fields. Or different minimums on different fields. Or something.

melinath:
Preventing wrapping altogether could lead to the opposite problem where fields with a lot of text end up being SUUUUUPER long

harrislapiroff:
Yeah, I mean, ideally they'd do the spreadsheet thing where they didn't wrap _and_ were cropped and you could click to view the whole value. (and maybe edit! grid editing!)

melinath:
Okay... I think that's getting waaay ahead of what we need right now.

Here's what I propose: we add a way for ModelTables to specify cell widths, either as keywords (auto / small / medium / large) or as widths (auto / 100 px / etc), based on the column's field name. Does that sound like a reasonable implementation?

melinath:
Would love to see this fixed. :-p

![screen shot 2015-09-15 at 6 03 58 pm](https://cloud.githubusercontent.com/assets/299979/9894067/5784a832-5bd4-11e5-8f09-7658331de4a9.png)

melinath:
I'm marking this as closed for now. I know we want a more thorough solution long-term, but we can make a new ticket if/when it comes to that point.

PR #672, opened 2015-10-21
Manually apply word-wrapping to cells with long values.

chigby:
What do people think about this as a simple solution to #473? 

The gist of it: if the value of a cell is long, wrap it using django's build-in word wrap filter, then turn off the table's auto-wrapping feature with `white-space: nowrap`, as we are already doing for iterable cells.

This has the advantage of not needing to change the ModelTable code or manually assign widths to specific columns.

The disadvantage, as I see it, is that we may end up having to do the above anyway if we ever want to change to `table-layout: fixed`, which is a CSS property I didn't know much about but does allow for more control over how the table is rendered. But you have to specify the column widths yourself rather than let the browser render them based on the table cell's contents. I have a feeling this process would be annoying, and especially so since these tables we're looking at don't have a fixed number of columns.

So, my hope is that this will tide us over until a full-on clickable & editable grid.

### DIFF
--- a/brambling/templates/brambling/event/organizer/__table.html
+++ b/brambling/templates/brambling/event/organizer/__table.html
@@ -159,7 +159,9 @@
                                         {{ cell|linebreaksbr }}
                                     </span>
                                 {% else %}
-                                    {{ cell }}
+                                    <span style="white-space:nowrap;">
+                                        {{ cell|wordwrap:60|linebreaksbr }}
+                                    </span>
                                 {% endif %}
                             {% endblock %}
                             {% if not forloop.first %}


melinath:
before & after with this branch. Definitely an improvement.

screen shot 2015-10-21 at 9 09 17 pm
screen shot 2015-10-21 at 9 08 58 pm

melinath:
What about just always running the cell through the `wordwrap:60` filter? Is there any particular reason not to do that?

chigby:
I think that makes sense. I didn't do that initially because I thought it might break things, but it seems fine.

PR #673, opened 2015-10-24
Order notes column

### LINKED ISSUE #437: Show order notes in the model table
following #436 (following #242)

melinath:
Resolved #437. Set the order notes column to display in the order table, and also moved the notes form to the front of the order detail page so it's easier to find & use.


melinath:
Went ahead and fully resolved #242 as well as part of this PR.

brambling/models.py
```diff
@@ -1459,6 +1459,9 @@ class Attendee(AbstractNamedModel):
 
     custom_data = GenericRelation('CustomFormEntry', content_type_field='related_ct', object_id_field='related_id')
 
+    # Admin-only data
```
harrislapiroff: I'm gonna change this to say "Organizer-only data". We're admins. Our customers are organizers.

melinath: okay

harrislapiroff:
The UI here is pretty wonky (confusing to distinguish between attendee notes and order notes; filling out multiple forms and pressing one save button only saves one form and loses all the other entered data) but I think I'm going to merge it because I'm pretty sure the best fix is to have the fields in the tables, which will require more significant changes.

harrislapiroff:
Actually maybe I'll clean it up a little.

harrislapiroff:
No wait, I can't. Not without restructuring the form. Eh. Too much. We'll fix it later.

PR #676, opened 2015-10-24
Set up proper utf-8 byte-encoding for csv downloads.

### LINKED ISSUE #643: Unicode chars (specifically the GBP symbol) make CSV export fail
Unicode encoding issues wooo. Works fine with the new excel export though. Must be an issue in the csv writer.

melinath:
Resolved #643.


chigby:
Wondering if this could also get the job done: https://github.com/jdunck/python-unicodecsv -- maybe then the CSV writer would "just work" like the Excel writer seems to.

Also wondering if it would be possible to get some test coverage on the CSV and XLS views? I tried to set up something as simple as humanly possible (i.e. making a GET request with `format=csv` and asserting a 200 response) but I think my understanding of Django testing and how the ModelTableViews do what they do is insufficient :(. Because it seems like the view is where the real problem is happening.

I'm happy to merge this because it does fix a currently-happening bug, but just wanted to know your thoughts on the above.

melinath:
Yeah, that's better. See #682 for that implementation + test.

PR #678, opened 2015-10-24
Made slug fields more intuitive. Resolved #496.

### LINKED ISSUE #496: Rename slug fields
Even with the help text, the name "Slug" is probably a little confusing and alienating to laymen. This applies to both events and organizations.

screenshot 2015-05-06 12 47 26

Ideally it could be called "URL" or "Web Address" or similar and the field could have the domain prefixed to it.

melinath:
Switched labels for event & org slug fields to "URL" and added a prepend to make clear what the final URL will be. Also added some help text to the org & event website/facebook fields.

### DIFF
--- a/brambling/templates/brambling/forms/event.html
+++ b/brambling/templates/brambling/forms/event.html
@@ -8,19 +8,21 @@
 	<div class="panel-body">
 		{% formrow form.name %}
 		{% if form.slug %}
-			{% formrow form.slug %}
+			{% formrow form.slug with prepend="https://dancerfly.com/"|add:organization.slug|add:"/" label="URL" %}
 		{% else %}
 			<div class="form-group">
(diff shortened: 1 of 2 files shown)


brambling/templates/brambling/forms/event.html
```diff
@@ -8,19 +8,21 @@ <h4 class="panel-title">Information</h4>
 	<div class="panel-body">
 		{% formrow form.name %}
 		{% if form.slug %}
-			{% formrow form.slug %}
+			{% formrow form.slug with prepend="https://dancerfly.com/"|add:organization.slug|add:"/" label="URL" %}
```
harrislapiroff: /me grumbles about hard-coding the domain and protocol into the template, then realizes it will probably never matter

harrislapiroff:
This looks awesome.

PR #680, opened 2015-10-25
Set up calendar prepends on date fields (and clocks on time fields.)

### LINKED ISSUE #491: Add calendar field addons to date fields
It would be nice to indicate visually that those fields have special behavior.

melinath:
Also tweaked discount form to make availability a little more obvious.
Resolved #491.


harrislapiroff:
This'll need to be fixed

screenshot 2015-11-03 15 16 06

Issue #661, opened 2015-10-10
Make pending payment amounts clear on attendee / order tables


melinath:
Right now, there's no easy way for organizers to tell whether there are pending payments on an attendee / order. We've had a request to make that more obvious.

melinath:
Related to #311

melinath:
This was resolved with the merge of #681.

PR #681, opened 2015-10-25
Pending payment amounts

melinath:
Split Attendee "paid" column and Order "balance" column into two columns for confirmed and pending transactions.


melinath:
Resolved #661

harrislapiroff:
Seems to work. I wonder if "Pending" should be "Owed" instead?

Issue #403, opened 2015-03-06
Clearer checkout flow when buying multiple transactions worth of items for an event 


harrislapiroff:
It can be confusing when purchasing items for an event because all of your previously purchased items show up alongside them.

Ideally I'd like this interface to work more like a shopping cart model, e.g., when you've completed a transaction, if you return to the shop, previous transactions do not appear during your next checkout process. It is an independent process. (I.e., when I buy a book from amazon, I don't also see all the other books I bought this month in my shopping cart.)

In practice, this will be infeasible or require a little more thought because a user may wish to assign newly purchased non-pass items to attendees whose passes have already been purchased.

An alternative design option would clarify on the attendee section of the checkout flow which items/passes are from previous transactions, either by marking them as purchased or by putting them in a separate section labeled something like, "You have previously registered the following attendees for this event."

melinath:
Related: #396

harrislapiroff:
Just received a complaint about this.

> XXXXXX XXXXXXXXXX should be emailing you shortly--she has successfully registered one person in her party and is having trouble registering the second person. The cart seems to be still displaying the first person's items and she can't remove them. Thanks for your help!

harrislapiroff:
The more I think about this the more I think that discrete labels (and adjusted colorings) indicating "purchased" vs "cart" is the right approach here.

harrislapiroff:
With attendee flow stuff resolved, this should be fairly easy. I think just adding a label on the item assignment page that says either "purchased" or "cart" will be clear. I'm therefore labelling this easy pickings.

PR #684, opened 2015-10-27
Attendee flow

### LINKED ISSUE #396: Item assignment UI unintuitive
See the discussion in #387.

melinath:
This is fully functional and pretty awesome. It could use some UI love – making it more obvious when something *can* be selected / *is* selected – which should probably happen as part of this branch.

This will resolve #396 and partially address #403 when it lands. At least that's the idea.


melinath:
(It also needs some explanatory text, which I would be happy to take a stab at or leave to someone else.)

brambling/static/brambling/attendees.ractive.js
```diff
@@ -0,0 +1,474 @@
...
+	function handleKeydown(event) {
+		if (event.which === 32) {
+			// space key
+			this.__tap_handler__.fire();
+		}
+	}
+
+	return ractive_events_tap;
+
+}));
```
harrislapiroff: Can we put libs in separate files?

melinath: Sure, I'll do that quick.

melinath:
The ractive shop seems to not be working quite right on this branch. Not sure if it's working on master, but... I'd kinda rather resolve that in a new branch since we're not using it yet anyway? *hopeful / sheepish grin?*

harrislapiroff:
Yeah, that sounds reasonable. I'll keep reviewing this.

harrislapiroff:
It would be preferable to me if this branch still auto-assigned items when only registering a single attendee. That's the most common case and should be as smooth as possible.

brambling/templates/brambling/event/order/_attendees_ractive.html
```diff
@@ -0,0 +1,90 @@
...
+					</li>
+				{{/if}}
+			</ul>
+			<div class='panel-footer'>
+				{{#if selectedItem}}
+					<span class='btn btn-default btn-block disabled'>
+						Return to shop for more
+					</span>
+				{{else}}
+					<a class='btn btn-default btn-block' href='../shop'>
```
harrislapiroff: This URL problem is going to get out of hand. We probably need to write a URL reversing API or something.

harrislapiroff: Not in this branch.

brambling/templates/brambling/event/order/attendee_basic_data.html
```diff
@@ -147,8 +164,16 @@ <h3 class="margin-leader-0">
 						</div>
 					</div>
 				{% endif %}
-				<button class='btn btn-primary' type="submit">{% if attendee.pk %}Save changes{% else %}Save and continue{% endif %}</button>
-				{% if attendee.pk %}<a href="../" class="btn btn-link">Cancel</a>{% endif %}
+				{% if attendee.pk %}
+					<button class='btn btn-primary' type="submit" name='next' value='review'>Save changes</button>
+					<a href="../" class="btn btn-link">Cancel</a>
```
harrislapiroff: This is a django template though and should definitely use `{% url %}`

brambling/templates/brambling/event/order/attendee_basic_data.html
```diff
@@ -147,8 +164,16 @@ <h3 class="margin-leader-0">
...
+				{% else %}
+					<button class='btn btn-primary' type="submit" name='next' value='review'>Continue</button>
+					<button class='btn btn-default' type="submit" name='next' value='add'>Register another attendee</button>
+					{% if attendees.count > 0 %}
+						<a href="../" class="btn btn-link">Cancel</a>
```
harrislapiroff: supra

melinath:
That should be doable. It would also be neat if one could move more than one item at once... Should be doable as long as they're all from the same source

harrislapiroff:
Does the masthead need to be ractive here? We need the countdown timer back on the attendee page.

harrislapiroff:
(Also I don't like making entire pages a single ractive instance, as previously discussed :p)

harrislapiroff:
Clicking the trash cans to delete attendees isn't working.

melinath:
I can fix the trash cans.

I guess the banner image doesn't have to be ractive, but the stepbar does. It has to react to whether the current page is finished or not.

brambling/templates/brambling/event/order/attendee_basic_data.html
melinath: supra?

harrislapiroff: see above

harrislapiroff:
Hm. I can see that being good, though I don't think it would be a big deal if the stepbar _didn't_ react since it doesn't react on other pages.

In my ideal world, the stepbar would be a completely separate component that would check the attendee component for completion. Especially since it will be reused on multiple pages ultimately. If necessary a parent component could coordinate that communication.

melinath:
yeah, I think having it as a separate component would be really cool.

harrislapiroff:
This looks pretty good actually. Great work on the overlays. Everything _could_ be spiffed up a bit, but I'm happy to get any of that and the other stuff I complained about after merging this. Feelings on merging as-is @melinath?

melinath:
Yep, I'm satisfied with the current state for now. :-)

harrislapiroff:
Let me give it one more pass.

harrislapiroff:
Actually, is it possible to bypass the item assignment page altogether if it's a single-attendee situation?

harrislapiroff:
I can take a swing at that.

melinath:
@harrislapiroff I don't think we should bypass it altogether. It's really easy to accidentally click the "Continue" button from attendee creation instead of "Register another attendee" – and it's not obvious how to go back and add another person if you intended to have more than one.

harrislapiroff:
I think it's very confusing (and a waste of time) to be presented with an interface that you don't have to do anything with and since single-attendee is the most common case, I'd prefer to optimize for that.

melinath:
Okay.

harrislapiroff:
If we continue to get confusion about registering multiple attendees, I'll reconsider. But right now, my instinct says that we'll get more confusion this way.

harrislapiroff:
@melinath Would you review that last commit and then merge if this is good?

melinath:
On it @harrislapiroff

PR #687, opened 2015-10-27
Can refund fix

### LINKED ISSUE #686: Transaction.can_refund always returns False if there was no money exchanged.
Right now, Transaction.can_refund looks for refund transactions and adds the sum of their amounts to the sum of the current transaction and returns true if that's over 0.

This means that transactions with an amount of 0 (for example full comps) will always return as fully refunded.

melinath:
Resolved #686 plus a bit. Basically, corrected Transaction refund-related behavior when Transaction's amount is 0 (for example, a fully-comped item). Includes new tests, oh boy!

### DIFF
--- a/brambling/models.py
+++ b/brambling/models.py
@@ -1234,13 +1234,18 @@ class Transaction(models.Model):
         if bought_items is None:
             bought_items = self.bought_items.all()
 
-        total_refunds = self.related_transaction_set.aggregate(Sum('amount'))['amount__sum'] or 0
-        refundable = self.amount + total_refunds
-        if refundable < amount:
-            raise ValueError("Not enough money available")
+        total_refunds = self.related_transaction_set.aggregate(Sum('amount'))['amount__sum']
+        if total_refunds is not None:
+            refundable = self.amount + total_refunds
+            if refundable == 0:
+                return None
 
-        if refundable == 0:
-            return None
+            # SB: Pretty sure this just disables completing partial
+            # refunds.
+            if refundable < amount:
+                raise ValueError("Not enough money available")
+        else:
+            refundable = self.amount
 
         refund_kwargs = {
             'order': self.order,
--- a/brambling/models.py
+++ b/brambling/models.py
@@ -1223,8 +1223,10 @@ class Transaction(models.Model):
     def can_refund(self):
         refunded = self.related_transaction_set.filter(
             transaction_type=Transaction.REFUND
-        ).aggregate(refunded=Sum('amount'))['refunded'] or 0
-        return self.amount + refunded > 0
+        ).aggregate(refunded=Sum('amount'))['refunded']
+        # None means there are no refunds, which is relevant
+        # for 0-amount transactions.
+        return refunded is None or self.amount + refunded > 0
 
     def refund(self, amount=None, bought_items=None, issuer=None, dwolla_pin=None):
         if amount is None:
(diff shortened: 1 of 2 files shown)


brambling/models.py
```diff
@@ -1223,8 +1223,10 @@ def from_dwolla_refund(cls, refund, related_transaction, **kwargs):
     def can_refund(self):
         refunded = self.related_transaction_set.filter(
             transaction_type=Transaction.REFUND
-        ).aggregate(refunded=Sum('amount'))['refunded'] or 0
-        return self.amount + refunded > 0
+        ).aggregate(refunded=Sum('amount'))['refunded']
+        # None means there are no refunds, which is relevant
+        # for 0-amount transactions.
+        return refunded is None or self.amount + refunded > 0
```
harrislapiroff: Seems a _little_ confusing for a function with a `can_` name to return a value that is neither `True` nor `False`.

melinath: ? This returns true or false.

harrislapiroff: or `None`

harrislapiroff: oHW AIT

harrislapiroff: JUST KIDDING HAHAAHHAHAHAHAHAHA

melinath: :-)

brambling/models.py
```diff
@@ -1232,13 +1234,18 @@ def refund(self, amount=None, bought_items=None, issuer=None, dwolla_pin=None):
...
+            refundable = self.amount + total_refunds
+            if refundable == 0:
+                return None
 
-        if refundable == 0:
-            return None
+            # SB: Pretty sure this just disables completing partial
+            # refunds.
+            if refundable < amount:
+                raise ValueError("Not enough money available")
```
harrislapiroff: Doesn't this (correctly) disable refunding more than the value of the transaction? A partial refund would be `refundable > amount` right?

brambling/tests/functional/test_transaction_model.py
```diff
@@ -0,0 +1,77 @@
...
+        self.assertEqual(txn.related_transaction_set.count(), 0)
+        refund = txn.refund()
+        self.assertIsInstance(refund, Transaction)
+        self.assertFalse(txn.can_refund())
+        self.assertEqual(txn.related_transaction_set.count(), 1)
+
+        refund = txn.refund()
+        self.assertIsNone(refund)
+        self.assertFalse(txn.can_refund())
+        self.assertEqual(txn.related_transaction_set.count(), 1)
```
harrislapiroff: You keep writing tests! :heart_eyes:

harrislapiroff:
This looks good and seems to function as expected, though I'd like to get some clarity on [this comment](https://github.com/littleweaver/django-brambling/pull/687#discussion-diff-43414961R1243). It can happen post-merge though.

brambling/models.py
melinath: Refundable is the amount that hasn't been refunded yet. (total_refunds is negative.) So it will never be more than amount. It'll be less if a refund of any kind has already happened.

harrislapiroff: Wait, isn't `amount` the amount that you are attempting to refund?

melinath: *facepalm* Yes. Yes it is.

harrislapiroff: hahaha, maybe commit a comment to avoid confusion in the future? :D

melinath: will do

PR #693, opened 2015-10-31
Hotfix/attendee duplicates

melinath:
Fixes duplication of boughtitems and attendees, with tests to prove it.


harrislapiroff:
Looks gr8

Issue #458, opened 2015-04-22
Event panel lists should limit number they display


harrislapiroff:
As organizers manage more events and attendees attend more events, these panels are going to become unwieldy.

We should limit the events displaying in these panels to future/current events and limit the number. We can provide a "See all events" link at the bottom which would take you to a separate view with a complete (paginated?) listing.

![screenshot 2015-04-21 22 48 36](https://cloud.githubusercontent.com/assets/296631/7266508/911bf89a-e878-11e4-84f2-45af91eb4ed9.png)

melinath:
I'm thinking more and more that these just shouldn't be visible from the homepage. I think "Your Events" should be "Browse Events", and that your orgs, events, and orders should be viewable on other pages.

For this ticket, I'd suggest just adding org / event pages (from the username dropdown menu). Order history would be a separate ticket. Side note: The org list should maybe list what payment methods are enabled and that should link to the org's payment settings.

harrislapiroff:
I'm -1 on removing from the homepage unless there's a good reason to. I think that if I'm an organizer, accessing my own event is the #1 task I'm going to be trying to do every time I visit Dancerfly and it should be immediately accessible.

The case for keeping events you're registered for on the homepage is a little less clear, but I also don't see a good case for making them harder to find.

Past events should definitely be removed. But ideally accessible _somewhere_.

melinath:
Part of my thought is also that it would free up space for a more advanced filter/search interface...

harrislapiroff:
I actually think accessing one's own events is more common than filtering and maybe filtering should be relegated to an interior page when logged in, but I actually think they can coexist. I'd prefer to see a thin horiz filtering interface like Eventbrite

![screenshot 2015-12-18 09 44 31](https://cloud.githubusercontent.com/assets/296631/11903137/fc866128-a56b-11e5-9a4d-43584e2e669e.png)

or OKC

![screenshot 2015-12-18 09 45 35](https://cloud.githubusercontent.com/assets/296631/11903158/1de122cc-a56c-11e5-9323-cb6a83af0213.png)

(also, kinda weird that the token shows up as "Monogamy" even if the actual search criteria is "Monogamous: No")

melinath:
Yeah, fair 'nuff.

PR #697, opened 2015-11-04
Refactored Organization detail view to show all events by the org, and added display of event privacy & publish state.

melinath:
Privacy is now displayed on list and detail pages for events; publish state is displayed on list view only, and only if user is an admin for the event.

The idea here was to prep for event duplication by making sure that organizers have a way to view *all* the events they might want to duplicate. I also think this display makes more sense for an org detail page generally.


melinath:
This at least partially addresses #458 as well.

brambling/templates/brambling/_event_list_item.html
```diff
@@ -13,7 +13,18 @@ <h5 class="datecard-month">{{ event.start_date|date:"M" }}</h5>
 			</div>
 		</div>
 	</a>
-	<h3>{{ event.name }}</h3>
+	<h3>
+		{{ event.name }}
+		{% if is_admin and not event.is_published %}
+			<small class='text-danger tipped' title="You can see this event because you're are an admin for it">
```
harrislapiroff: I'm gonna change from danger to something else. Danger should only be for things that you can fix or really should not do.

harrislapiroff: Actually. Lemme see how this looks.

brambling/templates/brambling/event/order/shop.html
```diff
@@ -179,32 +179,23 @@ <h4 id="myLargeModalLabel">{{ options.grouper.name }}</h4>
...
+									{% include "brambling/event/_privacy.html" %}
 								</div>
-								<div class="media text-large">
-									<div class="media-left">
-										<i class="fa fa-map-marker fa-fw"></i>
-									</div>
-									<div class="media-body">
-										{% include "brambling/event/_where.html" %}
-									</div>
+								<div class='margin-trailer-half'>
```
harrislapiroff: This whole bit is gonna require some visual/responsive review in a sec.

brambling/templates/brambling/_event_list_item.html
melinath: Someday later this should provide better instruction for how to publish.

Issue #299, opened 2015-01-09
Ability to duplicate an event


nmorduch:

melinath:
This is really closely tied to #301. It is slightly lower-priority.

melinath:
Once this is done, we should probably (in a separate ticket) remove the "event defaults" section & support. I think that path was a mistake... more confusing than helpful.

melinath:
I had been thinking of adding a "duplicate" button to the event list item, but I think it might actually be better to do something more like the saved attendee workflow – on the "Create an event" page, if they already have events, offer two choices: either copy an event or start from scratch. Copying an event presents them with a couple fields: choose event (Chosen field), new event name, new event start / end times. The rest they can edit later. Copy all the config over. If there are custom dates, use a diff between them and the old start date to determine the new date.

PR #698, opened 2015-11-04
Purged all references to event defaults.

melinath:
See #299. Event defaults adds a layer of complexity that would be better handled by letting people duplicate events.


melinath:
Actually, closing this. Will make this part of the duplication PR since there shouldn't be lag between the two.

PR #699, opened 2015-11-04
Copy events

### LINKED ISSUE #299: Ability to duplicate an event

melinath:
This PR resolves #299 (event duplication). I thought there was also a ticket related to making event creation use a different workflow than updating, but I can't find it now.

The event creation interface is pretty heavily inspired by [Github's new repo form](https://github.com/new).


brambling/templates/brambling/event/organizer/update.html
```diff
@@ -1,9 +1,295 @@
+{% load floppyforms %}
```
chigby: Pretty sure line being above the `extends` tag is causing the following problem for me:

```
TemplateSyntaxError at /demo/glodnom/edit/

<ExtendsNode: extends 'brambling/event/organizer/__narrow.html'> must be the first tag in the template.
```

harrislapiroff: There should probably be some integration tests that catch these sorts of things someday...

chigby: If I felt confident that I knew how to write integration tests in this framework, I'd totally write ones to make a request to each view and assert that it gives a 200 response during a Dancerfly sprint.

melinath: That would also be a good way to learn how to write tests in this framework.

harrislapiroff: Alternatively, I could do this too. It sounds fairly easy _and_ useful. I wonder about the best practice for structuring something like that though.

harrislapiroff:
```python
from django.test import TestCase
from django.test.client import Client

class MyViewTest(TestCase):

    def setUp(self):
        self.c = Client()

    def test_my_view(self):
        response = self.c.get('/path/to/view/')
        self.assertEqual(response.status_code, 200)
```

melinath: Essentially, but you'd also have to do prep data setup.

chigby: I think structure-wise it can just be one file per view? Or maybe file for each group of related views. This is also the place where you can put tests for any other basic stuff (e.g. assert the response is 403 if the user is not logged in, or logged in as the wrong user) or actual application logic happening in the view code itself.

brambling/forms/organizer.py
```diff
@@ -167,6 +152,116 @@ def save(self):
...
+            fields = (
+                'description', 'website_url', 'banner_image', 'city',
+                'state_or_province', 'country', 'timezone', 'currency',
+                'has_dances', 'has_classes', 'liability_waiver', 'privacy',
+                'collect_housing_data', 'collect_survey_data', 'cart_timeout',
+                'check_postmark_cutoff', 'transfers_allowed', 'facebook_url',
+            )
+            for field in fields:
+                setattr(instance, field, getattr(template, field))
+
```
chigby: I feel like we need to save the instance after we set all these attributes -- when I try it my template data isn't in the new event.

melinath: :-( but I tested it! With tests! :-( :-(

brambling/templates/brambling/event/organizer/create.html
```diff
@@ -0,0 +1,141 @@
...
+				$field.children().remove();
+				allEvents.each(function (idx, ele) {
+					var $ele = $(ele);
+					if ($ele.attr('value') == '' || $ele.data('org-id') == orgId) {
+						$field.append($ele);
+					}
+				});
+				$field.trigger("chosen:updated");
+			};
+		updateAvailableEvents();
```
chigby: By deleting and re-adding the available events here, it's breaking the "Copy Event" button. That is, when you visit the page with `template_event=N`, the event with id=N won't be initially selected because it's being removed from the select list and then added back (this time without the "selected" attribute).

One potential solution I see here is to just make the drop-down read-only if that GET argument is there. If they clicked the Copy Event button, then really they've already chosen what event they want as a template and don't need to mess with it anymore.

brambling/forms/organizer.py
melinath: Damn, yeah, the related objects are copied but not the core data, because it's not saved yet. :-p

chigby:
Other comments (which I know aren't strictly part of the changes you have made for copying events, but I'm putting them here anyway).

1. I love the URL entry field. 
2. I'd love the start/end date fields if the start date auto-filled into the end date (or if it combined both into a single "date range picker" thing like this: http://www.daterangepicker.com/)
3. On IE and Firefox the `<input type='time'>` doesn't exist and the users are told to input the time in 24-hour format. I feel like for users in the US this is clerical errors waiting to happen. Recommend using a time picker or parsing the time on the server (the python-dateutil package makes potentially easy).
4. I worry somewhat that the little X in the corner isn't an obvious enough way to select "No template."

brambling/forms/organizer.py
melinath: Okay, fixed.

melinath:
Glad you like the url field!

The time field issues are not unique to this branch and shouldn't be fixed here. Also, IE and Firefox? Holy Thorough Testing, Batman!

Similarly, the X is a feature of the Chosen fields we use and probably shouldn't be changed as part of this PR. The most I could do would be to make the field fit its content instead of being full-width.

I'll have a look at the date range picker...

melinath:
I have a quick autocompletion script for the event creation process, but I'm not 100% sure it's the behavior we want; I'm satisfied to merge as-is.

```
(function() {
	var $start_field = $('#id_start_date'),
		$end_field = $('#id_end_date'),
		end_modified = false;
	$start_field.on('change', function() {
		if (end_modified) return;
		$end_field.val($start_field.val());
	});
	$end_field.on('change', function() {
		end_modified = true;
	});
})();
```

PR #701, opened 2015-11-05
Lengthened dashboard to show all upcoming events, rather than just 5.

melinath:
I noticed that Rain City Blues can't be found on the dashboard. Turns out it's cause we limit the dashboard to five events because reasons. I don't think we need to do that thing any more.


harrislapiroff:
Pagination (probably [non-automatic] infinite scroll), someday, when we're wildly popular. Not necessary now, though.

PR #703, opened 2015-11-09
Updated a few requirements I'm confident won't hurt anything.

harrislapiroff:


harrislapiroff:
Feel free to update a few more on branch. This'll fix the build problem we're having with jdcal and I figured that as long as I was in there, I'd update a few other things. I was sort of random with what I chose to update though.

One note that's not random: we can't update grappelli to latest without updating Django to 1.8.

melinath:
Yep, doesn't seem to break anything.

PR #708, opened 2015-11-20
Org creation

melinath:
Set up org creation directly from event creation. I feel like there must be a ticket open for this but I can't find it right now.

Users without an org will only have the option to create. Other users default to existing org but can toggle between that and creating a new org.

### DIFF
--- a/brambling/forms/organizer.py
+++ b/brambling/forms/organizer.py
@@ -154,6 +154,9 @@ class OrganizationPaymentForm(forms.ModelForm):
 
 class EventCreateForm(forms.ModelForm):
     template_event = forms.ModelChoiceField(queryset=Event.objects.all(), required=False)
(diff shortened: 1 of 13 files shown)


harrislapiroff:
If City/State are not on the creation form, they shouldn't be required on the update form: 

screen shot 2015-10-03 at 5 23 30 pm

Otherwise the update form is unsubmittable by default.

melinath:
I'd be fine making them optional, I think?

harrislapiroff:
:+1:

harrislapiroff:
Seems to work. The UI isn't perfect, but it's good enough and cleaning it up will probably rely on some better ractive components/dropdowns. COOL!

harrislapiroff:
:cool: :sunglasses:

Issue #655, opened 2015-10-04
Hide housing data for folks who aren't requesting housing


melinath:
Via @harrislapiroff on #651:

> Not sure if this should be part of this PR, but is there any way to clear housing data if the user doesn't need housing. The use-case is that the attendee fills out the housing form, decides they don't need housing after all, then selects "Staying at own home"
>
>![screenshot 2015-10-04 17 10 50](https://cloud.githubusercontent.com/assets/296631/10270305/170f3b82-6abb-11e5-80c0-3f4956432edb.png)

I think we should still save what they've said, just not display it to admins.

melinath:
Resolved with merge #710

PR #710, opened 2015-12-15
Conditional housing data display

chigby:
One solution to #655

This way of doing it is fairly verbose, but I think it's also mostly transparent about what is happening and easy to change later. I can think of less repetitive ways of doing this, but there is already a lot of magic happening with these model tables. Trying to be explicit, here!


melinath:
This looks good in principle, but there are a few style changes that I'd like to see slash loose ends to clean up.

* I think I'd prefer to see a private `_display_housing_data` method on the model table rather than a public method on the model.
* This would be a good spot to use the python ternary operator - i.e. `return obj.other_needs if obj.needs_housing() else ''`
* The `related_objects_list()` wrapper exists as a thin wrapper around a manager - it's just shorthand for a method calling obj.related.all() so that it takes one line instead of two. Calling it *from* a defined method defeats the little benefit it added and adds a layer of distance that could end up being confusing later. And really the layer of distance it already adds is not great. Probably best to just get rid of it.
* We should handle the reverse case while we're at it – i.e. if someone had said they'd host but changed their mind.

chigby:
I can go ahead and make these changes -- they are good! Though do you think that `Attendee.needs_housing` is unnecessary? To me it's reasonable that an Attendee knows whether it needs housing and the model table doesn't need to know exactly how that's implemented. `_display_housing_data` would call `attendee.needs_housing`.

melinath:
I'm pretty ambivalent about `Attendee.needs_housing()`... I'm fine with it staying in.

melinath:
(Also, if you do have needs_housing() the _display_housing_data is of course unnecessary.)

chigby:
I guess my general philosophy is that the cost of adding methods to models is pretty low. Would you directly compare against `Attendee.NEED` (or even the string `'need'`) in the model table?

melinath:
Adding methods is in principal low-cost, but it can easily make the code less clear (slash explicit) by adding levels of indirection. For a simple shortcut like this, it also means there's more than one obvious way to do things. These both mean that people coming on to the project in the future will have a tiny bit higher of a learning curve – and the code will be less immediately understandable to us as well, if we leave it alone for a while. And the benefit doesn't seem to be that great – it doesn't make a common pattern more compact, it just makes a one-line comparison a bit shorter.

That being said, `needs_housing()` is a _very_ semantic method name, and Attendee doesn't have a lot of methods for it to get lost among, so I think it would probably be fine. It might even be a bit more understandable than `attendee.housing_status == Attendee.NEED`...

Hence: ambivalent in this case.

melinath:
@chigby is this ready for review again?

brambling/utils/model_tables.py
```diff
@@ -615,7 +638,7 @@ def get_field_val(self, obj, key):
             field = "spaces"
 
         if date_str:
-            if obj.get_eventhousing():
+            if obj.get_eventhousing() and obj.providing_housing:
```
melinath: `if obj.providing_housing` should be moved up a couple logical levels. It needs to also wrap the following section about custom housing data.

brambling/utils/model_tables.py
```diff
@@ -399,9 +399,10 @@ class AttendeeTable(CustomDataTable):
         ('Contact',
          ('email', 'phone')),
         ('Housing',
-         ('housing_status', 'housing_nights', 'housing_preferences',
-          'environment_avoid', 'environment_cause', 'person_prefer',
-          'person_avoid', 'other_needs')),
+         ('housing_status', 'housing_nights_if_needed',
+          'housing_preferences_if_needed', 'environment_avoid_if_needed',
+          'environment_cause_if_needed', 'person_prefer_if_needed',
+          'person_avoid_if_needed', 'other_needs_if_needed')),
```
melinath: The field names should stay the same, since we don't need to distinguish between if_needed and if not needed. Simpler change. Also less likely to have unexpected side effects... in this case, for example, the old field names are still being used to determine what data to prefetch.

brambling/utils/model_tables.py
```diff
@@ -506,6 +507,28 @@ def confirmed(self, obj):
     environment_avoid = related_objects_list("ef_avoid")
     environment_cause = related_objects_list("ef_cause")
```
melinath: You should be able to remove these related_objects_list lines now that it's not being used below.

brambling/utils/model_tables.py
```diff
@@ -734,7 +757,7 @@ def person_avoid(self, obj):
 
     def get_eventhousing_csm(self, obj, name):
         eventhousing = obj.get_eventhousing()
-        if eventhousing:
+        if eventhousing and obj.providing_housing:
             return related_objects_list(name)(self, eventhousing)
```
melinath: Any chance this related_objects_list (and the function in general) could also get replaced while we're in here?

brambling/utils/model_tables.py
chigby: I've been playing with this a little and it seems like something else is happening in that other block where if it doesn't happen I get errors like this:

`'Order' object has no attribute 'hosting_spaces_20160731'`

melinath: hmm. yeah, that kinda makes sense... because it doesn't try the date_str path, it tries the default: getting the attribute.

melinath: Probably needs to return '' instead of just getting skipped.

chigby:
@melinath Okay, this should be good to review again.

melinath:
@chigby failing tests on postgres

chigby:
Is there a good way to run the tests locally on postgres?

melinath:
If you set up a postgres server and then just update the DATABASE dict locally to point there instead of sqlite, it should work fine.

melinath:
Slash... setting DB=postgres in the env should trigger the test settings already included in settings.py

chigby:
@melinath: Ok, _now_ it's working (curse my hard-coded primary key references).

melinath:
hard-coded pk references always get you in the end! :-) That was like 90% of the issues I had switching MR tests over to mysql.

brambling/utils/model_tables.py
```diff
@@ -374,11 +360,19 @@ def get_custom_fields(self):
...
+        else:
+            return False
+
     def get_field_val(self, obj, key):
         if key.startswith('custom_'):
             if not hasattr(obj, '_custom_data'):
                 raw_data = {
-                    entry.form_field_id: entry.get_value()
+                    entry.form_field_id: (entry.get_value() if self._show_housing_data(obj, entry) else '')
                     for entry in obj.custom_data.all()
```
melinath: This should be `obj.custom_data.select_related('form_field__form'), otherwise it'll be 2nm additional queries (where n is the table size and m is the number of custom fields).

melinath:
Hm, it looks like this doesn't actually handle the hosting side of things - but that works out fine for now because the forms already (currently) delete the hosting information when you uncheck the box. (And I don't expect that to really change in the future.)

brambling/utils/model_tables.py
```diff
@@ -374,11 +360,19 @@ def get_custom_fields(self):
     def _get_custom_fields(self):
         raise NotImplementedError
 
+    def _show_housing_data(self, attendee, form_entry):
```
melinath: This doesn't seem like it should be on CustomDataTable, since it only actually has an effect on the AttendeeTable. In fact it's a little misleading because the second arg will only be an attendee if this *is* an attendee table.

Maybe a _get_custom_data method instead, which returns what becomes raw_data? Then AttendeeTable could override to add this filter, and OrderTable could override to add the data from eventhousing.

chigby:
No? There's a couple of tests for displaying the hosting information on the order table. Am I testing the wrong thing?

melinath:
This doesn't display things if the box is unchecked, yes, but as far as I can tell that's because the data is deleted, not because it isn't displayed. I can't test manually because there's no way to end up in that state.

melinath:
This is why line 647 of model_tables.py doesn't need to look like line 375. (This refers specifically to custom data fields. The date-related fields *are* handled.

Again though, this isn't a place where anything needs to change, just an interesting fact.

chigby:
Okay, I think this addresses all of @melinath's comments now.

melinath:
I found an edge case that still isn't handled, but it's really more an issue with the system as a whole rather than an issue with this PR. See #718 for more details.

PR #711, opened 2015-12-16
Added form handling of ValueErrors that might be raised while attempting to refresh dwolla information.

melinath:
I am fairly certain that these ValueErrors (caused by expired dwolla refresh tokens) could still crop up if a user's token was expired (rather than the organization's) but this at least plugs the one hole (during validation)

### DIFF
--- a/brambling/tests/functional/test_order_forms.py
+++ b/brambling/tests/functional/test_order_forms.py
@@ -1,7 +1,9 @@
 from decimal import Decimal
+from datetime import timedelta
 
 from django.test import TestCase
-from mock import patch, call
+from django.utils import timezone
+from mock import patch
 import stripe
 
 from brambling.forms.orders import (OneTimePaymentForm,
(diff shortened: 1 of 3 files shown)


brambling/tests/functional/test_order_forms.py
```diff
@@ -219,6 +218,20 @@ def test_dwolla_payment_form(self, dwolla_charge, dwolla_get_sources):
         self.assertEqual(Decimal(str(txn.application_fee)), Decimal('1.05'))
         self.assertEqual(Decimal(str(txn.processing_fee)), Decimal('0.25'))
 
+    @patch('brambling.forms.orders.dwolla_get_sources')
+    @patch('brambling.forms.orders.dwolla_charge', side_effect=ValueError('this is an error'))
```
chigby: This seems slightly artificial to me. I'd prefer something that mimics the error this code change is meant to handle more exactly, e.g. (omitting unchanged lines)

```python
@patch('dwolla.oauth.refresh')
def test_dwolla_payment_form_handles_valueerror(self, refresh, dwolla_get_sources):
    refresh.return_value = {u'error': u'access_denied', u'error_description': u'Invalid refresh token.'}
```

melinath:
@chigby fixed!

Issue #714, opened 2015-12-17
Connecting the same dwolla account to your org and your account / order invalidates the other's refresh token


melinath:
I.e. if I connect my user account to dwolla, then go connect my org, my user account's credentials will no longer be valid - including the refresh token. And vice versa. This is not good. Especially since we need different app permissions for each of them.

melinath:
Offhand, the best solution would seem to be creating a DwollaAccount model that consolidates each account's most current tokens in one place. It would need to also handle consolidating permissions... is that even possible? When you're connecting a new dwolla account, there's no way for *us* to know what it is (and thus check it against our database) until *after* the connection is finished. And by that time you've already potentially lost the other permissions we need from you.

(Customers need to give Funding permission so we can pay orgs on their behalf; orgs need to give Transactions permissions so we can get the org's transaction id (see dwolla_charge). Maybe we should just request all the permissions from everyone?

melinath:
[ ] Add DwollaAccount model
[ ] Add migration of old data to the new model
[ ] Combine oauth url generation / standardize on full permissions
[ ] Views should still use old dwolla credentials as possible, but re-connecting should be simplified.
[ ] Users should get alerts about their orgs / accounts with outstanding need to upgrade.

melinath:
Resolved with PR #716.

PR #716, opened 2015-12-22
Better Dwolla account handling

melinath:
This PR separates the dwolla account info off into a separate model, which is only *referenced* by orders, people, and orgs. This allows for centralized updates of access tokens related to a particular account, and prevents a bug whereby attaching both your org and your account to the same dwolla account results in access token invalidation. (See #714, which this PR resolves.)

This PR transfers dwolla account info for organizations over to the new model, but encourages them to update as soon as possible. (Though that isn't, strictly speaking, necessary... the access tokens would be properly updated and re-scoped the next time they attach their user account or an order. Maybe we just shouldn't say anything?)

This PR doesn't remove the old dwolla fields from the respective models, but it does remove them from the admin. If these changes are successful, a followup PR should remove the old fields and the data stored in them.


chigby:
This generally looks good, though the VCR-enabled test would only pass if I removed the `../` from the beginning of the path. I went through the reauthorization flow and everything seemed to check out for the testing dwolla account. 

Maybe others want to weigh in on whether to display the message?

melinath:
I think that in the light of day, it just doesn't make sense to ask them to re-authorize. The old tokens do everything orgs need, and as soon as someone authorizes for their personal account (or another org, or whatever) the scope will be expanded automatically.

Issue #607, opened 2015-09-12
Send receipts per-transaction rather than per-order


melinath:
In other words, when you make a transaction, you should get an email for *just* that transaction – how much you paid and what you bought – not for every transaction / item associated with that order.

harrislapiroff:
:+1:

chigby:
The receipt emails make a few references to the order code. Should we just use the code from the order that the transaction belongs to, add a per-transaction code, or something else?  I lean towards the first option but I don't know if people have strong feelings about this.
@harrislapiroff @melinath

melinath:
@chigby use the code from the order the transaction belongs to.

harrislapiroff:
👍

chigby:
Should this also apply to the Order Alert emails? Are the expectations of event organizers different from the person actually spending the money?
@melinath

melinath:
@chigby hmm, good question. I think that probably the order alerts to organizers should also be per-transaction, yes.

PR #719, opened 2015-12-23
Per transaction receipts

chigby:
Addresses #607.

Updates the order receipt mailer and the order alert mailer (which go to the attendee and the event management, respectively) to list the bought items and total amount of money involved for that transaction only, as opposed to all transactions for the same order. I did not change the design of the emails, just where they get their data from.


melinath:
I think the design for the order receipt is fine, but we should maybe update the copy slightly for the order alert - it's a little weird to have the email say 'you have a new order' when in fact it might be an order they've already seen... but I'm not immediately sure what would be better.

melinath:
@chigby if this all looks good to you, feel free to merge.

PR #722, opened 2015-12-24
Markdown support on event description

### LINKED ISSUE #441: Event/Item Description Formatting
It might be nice to give organizers some minimal formatting options for to avoid unformatted blocks like:

screenshot 2015-04-01 13 39 54

Maybe via a WYSIWYG editor like CKEditor?

melinath:
This resolves #441 (event description formatting) by adding HTML removal (with bleach) and markdown syntax parsing (with python-markdown) to the event description field. This also adds a reusable include that provides a modal explaining how to use markdown on dancerfly.


melinath:
@harrislapiroff if you could have a look at this that would be great. I know you probably will have feelings about it.

harrislapiroff:
Using markdown doesn't resolve #441 IMO. I think markdown is great for people who have a good degree of computer-comfort, but I rarely see it used outside of situations where users are mostly tech people. I think a simple WYSIWYG is superior (and slicker) for most people. This could be an intermediate step or an optional alternative though.

harrislapiroff:
That's my feelings review. I can also do a technical review later this week or next, since I'd be happy to merge this as an intermediate step, keeping #441 open.

melinath:
I'm happy to have us pursue a different path. I think we can expect a *little* more effort / knowledge from organizers, which might make this work, (and which I think used to be your argument for it?) But I'm also 100% down with doing a rich text editor instead.

melinath:
@harrislapiroff I've added tinymce (much less painful than ckeditor IMO and still pretty) on top of the markdown. I.e. this is still getting *stored* as markdown, but we provide the user with a rich text interface by default, with the option to edit source if they want.

brambling/templates/brambling/event/organizer/update.html
```diff
@@ -295,6 +296,8 @@ <h4 class="panel-title">Payment</h4>
 
 {% block javascripts %}
 	{{ block.super }}
+	<script type="text/javascript" src="{% static 'zenaida/js/bootstrap/tab.js' %}"></script>
```
harrislapiroff: Is this for anything?

harrislapiroff: Wait nvm. I see.

harrislapiroff:
We should disallow markdown images `image` works fine right now and is a slight security risk. (Someone could use an image to capture request data from their attendees and potential attendees.)

harrislapiroff:
I don't think the bottom bar with the HTML hierarchy is necessary.

test_project/requirements.txt
```diff
@@ -22,6 +22,9 @@ djangorestframework==3.3.0
 openpyxl==2.2.6
 jdcal==1.0.1
 unicodecsv==0.14.1
+markdown==2.6.5
+bleach==1.4.2
+html5lib==0.9999999
```
harrislapiroff: Seriously? O_O

brambling/templates/brambling/event/organizer/update.html
melinath: I think it used to be for a thing which was then removed. Which is maybe what you saw.

test_project/requirements.txt
melinath: Yep!

brambling/templates/brambling/event/organizer/update.html
harrislapiroff: So can I remove this?

brambling/templates/brambling/forms/markdown/_js.html
```diff
@@ -0,0 +1,90 @@
+<script src="https://cdn.tinymce.com/4/tinymce.js"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/to-markdown/1.3.0/to-markdown.js"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/markdown-it/5.1.0/markdown-it.js"></script>
+<script>
+	$('[data-toggle="tab"]').click(function (e) {
+		e.preventDefault();
+		$(this).tab('show');
+	});
```
harrislapiroff: And remove this?

melinath: I'm pretty sure, yeah. It was the old method of switching between WYSIWYG & markdown - but tinymce supersedes that.

harrislapiroff:
Refreshing changes the markdown to HTML.

harrislapiroff:
screenshot 2016-01-21 20 50 00

PR #723, opened 2015-12-24
Shorter admin sidebars

### LINKED ISSUE #458: Event panel lists should limit number they display

melinath:
I've shortened the sidebars and added simple separate views for display full lists of events & orgs you're an admin on. This resolves #458. I've also made orgs sort by last_modified and set up events & orgs to also update their last_modified dates when certain related objects are updated.


harrislapiroff:
This looks neat! Will try to review soon.

PR #725, opened 2015-12-24
Added basic notification settings (including laying the groundwork for dancerfly status update emails.)

### LINKED ISSUE #343: Notification settings
It might be good to let people control what kind of emails they receive. Organizers may not want to get 200 "you got an order" emails.

melinath:
With this PR, users get a notifications settings page which allows them to opt out of organizational alert emails. Future changes (either in this ticket or separately) once PR #719 is merged:

* [x] Add test for whether order alert sending actually obeys notify_new_person directives (I'm sure it does; the code is simple. But we should still test it.)
* [x] Add note to order alert email telling people how to change their notification settings / linking to them.
* [x] Add support for daily digest instead of just every email.

The checkboxes above, if checked, mean fixed here or new issue created.

Resolved #343.


melinath:
Thank goodness for CI.

chigby:
A couple of things to think about.

1. Separating the concept of "notification" from "email." I don't know if this is a planned feature, but I could imagine having some kind of organizer-based or event-based feed than consumes notifications generated elsewhere that works in parallel to the emails we're already sending. Which is a complicated feature, but maybe what I'm really saying here is that the db columns should be called "email" instead of "notify." Just in case.

2. I'm actually sort of wary of throwing boolean settings onto database tables, as things can get out of hand pretty easily, but I think we probably fine for now.

3. If we're concerned about people being flooded with emails, as in #343, then the system wherein we give people the choice of altering their settings _after_ they already got flooded doesn't feel that great. Also, the only choice is opting out of all emails entirely. I feel like what this issue is asking for is more like an option to get a daily digest email (or, like, a daily order summary, e.g. "you got 40 orders"). This could be part of the sign up process.

melinath:
1. We could have an internal notification mechanism at some point in the future, but if we do, it would be trivial to rename the db fields at that point.
2. Yeah, this might fit better as a SET field, but Django doesn't have built-in support for those. Hypothetically we could install a third-party package that adds that support... but then that would make the form / template side of things much more complex, and it would make non-binary options hard to do...
3. On the one hand, I agree. On the other hand, if we *don't* send them the first order email(s), they'll never even know that getting emails is an option. I think defaulting to "on" is not unusual or terrible, though in this case it could mean getting a large number of emails before you get a chance to change your settings. I think that, though it's more complicated, implementing a digest and defaulting to that is probably the way to go.

chigby:
Yeah, I agree it should default to true. I was more saying it should maybe be something they choose when they sign up. But it's also probably fine for now since it's how the site works currently anyway.

chigby:
I feel like the main barriers to a digest system are using some kind of job-runner. Do we have that?

Also the email template could potentially be cumbersome and probably needs some design love.

melinath:
I'm also hesitant to have it as part of the signup process because the majority of people signing up *won't* be running events. If anything I'd want it as part of their first event setup. Part of me even wants to offer per-event settings... but I think that's waaay too complicated for now, both implementation-wise and in terms of giving people a good overview of their notification options.

Implementing the digests would be a matter of adding a management command that sent out all relevant notifications and then running it as a cron job. Long-term we'll probably want this to go through celery, but for now cron should be fine. I guess we'd probably set up a 'notification sent' model that tracks when the last send was, and each time we'd send everything that hadn't been sent before. ... actually we'd just need a field on the Person model that gets updated each time the notification is sent.

chigby:
Yeah, I think the simplest method would be having `last_emailed_at` field on the Person model. If we really want to get into customizations, I could also potentially see a numeric field for how often they want the digest, every 1 day, 3 days, 7 days, etc.

It _is_ a little bit complicated by the fact that really we do want these just to apply to organizers. Probably needs more discussion.

I'm happy to merge this branch now and think about this stuff as part of the process of adding status update emails and so forth.

melinath:
I don't think we need a numeric field – I think maybe every transaction / 3-hour summaries / 1-day summaries / no emails is probably a good starting set. I could also see adding a 3-day / 7-day *maybe*, but definitely nothing beyond that, and even those seem of limited utility to me. Point being: limited number of options to cover the likely use cases, rather than something fully-customizable.

I think adding a daily digest would be fine as part of this PR; it's not super urgent. I can work on that at some point. (And it would make more sense here than with the status update emails. And it would mean we wouldn't have to set up a data migration later.)

melinath:
This is ready for review again.

melinath:
@chigby looks like this one is assigned to you! I forgot.

brambling/tests/functional/test_daily_digest.py
```diff
@@ -0,0 +1,115 @@
...
+from brambling.models import Person, Transaction
+from brambling.tests.factories import (
+    PersonFactory,
+    OrganizationFactory,
+    EventFactory,
+    TransactionFactory,
+)
+
+
+class DailyDigestCommandTestCase(TestCase):
```
chigby: Wondering if there should be a test here for the cutoff time aspect of this mailer.

brambling/mail.py
```diff
@@ -151,10 +154,42 @@ def get_recipients(self):
...
+
+    def __init__(self, recipient, cutoff, *args, **kwargs):
+        self.recipient = recipient
+        self.cutoff = cutoff
+        from brambling.models import Transaction
+        self.transactions = Transaction.objects.filter(
+            Q(event__additional_editors=recipient) |
+            Q(event__organization__editors=recipient) |
+            Q(event__organization__owner=recipient),
+            timestamp__gte=cutoff,
```
chigby: As in, we are testing the recipient aspects of this query but not the cutoff aspect.

brambling/management/commands/send_daily_emails.py
```diff
@@ -0,0 +1,41 @@
...
+from django.utils import timezone
+
+from brambling.mail import DailyDigestMailer
+from brambling.models import Person
+
+
+class Command(BaseCommand):
+    def get_recipients(self):
+        return Person.objects.filter(
+            notify_new_purchases=Person.NOTIFY_DAILY,
```
chigby: Do we think there should be a filter here for `last_new_purchases_digest_sent`? I sort of trust our code more than I trust cron, and it would let us safely run the script as often as needed without overnotifying people.

chigby:
@melinath That's all I have to say about this. The email itself looks good, as does the notification settings page.

brambling/management/commands/send_daily_emails.py
melinath: That filtering happens as part of the Mailer. It only gets transactions that happened since last_new_purchases_digest_sent (up to a max of 1 day). This is where the cutoff in this command's send_digest() method comes in.

chigby: Sorry if my comment was a little vague. I think what I'm saying is that there is nothing inherently daily about the send daily emails command. Right now it looks like (and correct me if I'm wrong) that the command will send emails about transactions that have happened since the last time the command was run, and cron is the thing that ensures it happens only once per day.

Whereas if there were a filter here to only select, e.g., people with `last_new_purchases_digest_sent` older than 24 hours, there's no chance of anyone being emailed their daily digest twice in that interval.

melinath: Mm. Though that could result in someone only getting an email every other day. Maybe a compromise like 20 hours?

chigby: I mean, we could have the cron job run every 2 hours as long as we only email people once per day.

melinath: True, but then we would have the problem of people getting emailed two hours later each day, which is weird. Down the road I could see that being a good method if we let people individually choose when their daily digest is sent. But for now I'd rather just stick to emailing everyone at the same time.

PR #730, opened 2016-01-04
Fix unconfirmed check detection in order receipt emails.

chigby:
I think as part of #719 the `unconfirmed_check_payments` flag became lost. This brings it back.


melinath:
Good catch, looks good.

PR #731, opened 2016-01-05
Name rename

JCattera:
Now displays "First Name" and "Last Name" instead of "Given Name" and "Surname". Also changed abbreviations for name ordering from "GMS" to "FML" and so on.


harrislapiroff:
So, migration `0044_auto_20160105_1929.py` is not quite right. You need to think about this a little differently. Here's how I'd advise you to proceed:

1. If you've already migrated, run it backward by running `./manage.py migrate brambling 0043`.
2. Delete the `0044_auto_20160105_1929.py` file.
3. Create a new migration with `./manage.py makemigrations`. This time, just leave this migration as is.
4. Create *another* empty migration with `./manage.py makemigrations brambling --empty`. (I am revising what I advised you on earlier, just because it's often cleaner to keep data migrations separate from the automatic ones.)

Migrations consist of a series of operations that contain both forward and reverse actions. These are defined in a `Migration`s `operations` attribute. `RunPython` is basically a way of writing your own custom migration type that's different from the built-in ones like `AlterField`, `AddField`, etc. You'll want to edit the empty migration. That will look something like this:

```
# -*- coding: utf-8 -*-
from __future__ import unicode_literals

from django.db import migrations, models

def forwards_func(apps, schema_editor):
    Attendee = apps.get_model('brambling', 'Attendee')
    # ... code goes here to change all GMS to FML etc. Those lines will probably look
    # something like `Attendee.filter(....).update(....)`
    # Refer to the QuerySet API: https://docs.djangoproject.com/en/1.9/ref/models/querysets/
    # You will need to do this for every affected model.

def reverse_func(apps, schema_editor):
    # Same thing, but backwards.

class Migration(migrations.Migration):

    dependencies = []

    operations = [
        migrations.RunPython(forwards_func, reverse_func),
    ]
```

harrislapiroff:
https://github.com/littleweaver/django-brambling/blob/master/brambling/migrations/0005_copy_check_postmark_cutoff.py is an example of a data migration in brambling. But that one doesn't specify a reverse migration.

PR #732, opened 2016-01-05
Name rename

JCattera:
Now displays "First Name" and "Last Name" instead of "Given Name" and "Surname". Also changed abbreviations for name ordering from "GMS" to "FML" and so on.


chigby:
Everything seems good here for renaming the fields.

I wonder if there's any way to simplify the migrations a little bit. In particular, it seems like number 42 is undoing a lot of work, and that's being redone in number 43. @melinath is there a standard procedure for this?

melinath:
Yeah, the migrations are a mess. It looks like you could delete 0041 and 0042 and then renumber the other two & hook the *new* 0041 up to the same dependencies that the *old* 0041 had.

brambling/admin/forms.py
```diff
@@ -42,7 +42,7 @@ class PersonChangeForm(forms.ModelForm):
 
     class Meta:
         model = Person
-        fields = ('email', 'password', 'is_active',)
+        fields = ('first_name','last_name','email', 'password', 'is_active',)
```
melinath: These shouldn't need to be added here, since they weren't mentioned before either. Are you sure they should be here? If not, we should probably not add them. (Also, code style note: always put a space after a comma.)

JCattera: that was probably a mistake as well. will revert it.

melinath: Kk. Once this is reverted I'll merge the PR. It's possible that there are instances of given_name and surname that we've missed, but this seems pretty comprehensive (and it's not easy to find things we might've missed without merging these changes first.)

PR #734, opened 2016-01-05
Added note if a user's upcoming event is awaiting check confirmation.

### LINKED ISSUE #499: Unpaid events by check show up as registered
Currently if you do payment by check for an event, the front page will say that you're registered. While this is _acceptable_ since it's technically correct, if the check is not yet marked as received, it would probably be better for it to display the "almost registered" button with a note about completing payment.

chigby:
Fixes #499 

This pull request makes it so that if a user is looking at their list of upcoming events (on the home page), and there's an event they've paid for by mailing in a check, then it will say "Almost complete" with a little box explaining that the organizer has yet to confirm the check payment. The wording in the box is taken from the receipt email, but can be anything, really.


brambling/templates/brambling/_event_list_item.html
```diff
@@ -40,6 +40,10 @@ <h5 class="text-muted event-tags margin-0">{% for style in event.dance_styles.al
 					<a href="{{ event.get_absolute_url }}" class="margin-trailer-half btn btn-warning tipped" title="Complete your registration now!">
 						<i class="fa fa-exclamation-circle"></i> Almost Registered
 					</a>
+				{% elif event.order and event.order.awaiting_check_confirmation %}
+					<a href="{{ event.get_absolute_url }}" class="margin-trailer-half btn btn-warning tipped" title="zokn okn">
```
melinath: I think "zokn okn" might not  be the most helpful *title=* here. ;-)

brambling/models.py
```diff
@@ -1098,6 +1098,10 @@ def get_eventhousing(self):
     def has_dwolla_payments(self):
         return self.transactions.filter(method=Transaction.DWOLLA).exists()
 
+    def awaiting_check_confirmation(self):
```
melinath: This method concerns me, because it hides what's actually happening (running a query) and makes it easy to call this method without noticing the cost. So for example, this method currently gets called once or twice for each event on a user's dashboard. I'd rather see this as a single query to get up to one unconfirmed check per event in the list, then pass that to the context. Easiest might be to give the context a list of tuples `(event, order, has_unconfirmed_check_payments)` rather than a queryset of events... not sure.

melinath: (side note: in general we try to stick to a convention of using get_<whatever:> for methods that perform database queries or other expensive calculation.)

melinath: (Other side note, `has_cart()`, which you were probably basing this on, clearly doesn't follow that convention either :-p and has the same problem. Maybe that could be fixed as part of this too?)

melinath:
The note beneath the event section feels really unwieldy to me, but I'm not sure I have a better solution. My inclination would be to use the tooltip text area for the note – but then an incomplete check payment would look the same as an incomplete registration, which would be confusing. Maybe change the text for the check payment to "Check pending" or something?

brambling/templates/brambling/_event_list_item.html
harrislapiroff: I think this tooltip needs a tooltip

brambling/models.py
chigby: It was modeled after `has_dwolla_payments`, which is in a similar situation.

melinath:
@harrislapiroff I'm pretty sure this is awaiting a design eye before work on it proceeds.

brambling/models.py
harrislapiroff: @melinath Was this resolved to your satisfaction?

brambling/views/core.py
```diff
@@ -47,7 +47,8 @@ def get_context_data(self):
             re_dict = dict((e.pk, e) for e in registered_events)
             orders = Order.objects.filter(
                 person=user,
-                bought_items__status=BoughtItem.RESERVED,
+                bought_items__status__in=(BoughtItem.BOUGHT,
+                                          BoughtItem.RESERVED),
```
harrislapiroff: Whaaaaaaaaat is the difference between `BoughtItem.BOUGHT` and `BoughtItem.RESERVED`????

harrislapiroff: Not really a query about this PR, but I don't know what this change actually does without understanding that.

melinath: `BOUGHT` is something that has been paid for; `RESERVED` is in the cart (and potentially expiring); `UNPAID` is a special status that counts as in the cart but doesn't get expired.

brambling/models.py
melinath: I... can't tell immediately. @chigby rebased, so I can't see if there were commits after this. :-/

brambling/views/core.py
harrislapiroff: Ohhhh I see what's happening here. Weird.

brambling/models.py
harrislapiroff: It looks like all the things you said in your comment still apply to this method, so I guess the question is more if it is resolved in your mind than if it is resolved in the code.

melinath: Yeah... it would be great if this could get fixed, since it adds one or two queries per event on the dashboard. It should be possible to add a qs of unconfirmed check payments to each event on the dashboard using a [custom Prefetch object](https://docs.djangoproject.com/en/1.9/ref/models/querysets/#django.db.models.Prefetch). Ideally you'd add a count or an exists rather than a full qs, but I don't think that's an option...

harrislapiroff: Okay, cool. In that case I'm bouncing this PR back to @chigby for more work. I could also give it a swing, but I think @chigby is still less busy than me. :stuck_out_tongue_closed_eyes:

melinath: Though OTOH it only kicks in on items with orders, so maybe that's not soooo bad...

harrislapiroff:
@chigby 

- [ ] Resolve [Stephen's issue](https://github.com/littleweaver/django-brambling/pull/734#discussion-diff-49044389R1101)

brambling/models.py
melinath: I think I'd be fine with it if the method started with `has_` to match `has_dwolla_payments`. That would be the start of a convention / something we could see and know needed to be fixed?

chigby: I feel like I started writing a better version of this, but it got lost in the mists of time, I'll see if I can find where I left off.

chigby:
@melinath What do you think of the method here 40e13dd?

melinath:
@chigby calling .count() on the querysets will trigger additional queries. Just check whether they evaluate as truthy themselves.

melinath:
It would also be nice to remove the merge commit of the branch into itself from the history...

chigby:
@melinath I think this is good now

PR #736, opened 2016-01-08
Subject encoding

JCattera:
Fixed the subject.txt files in the mail templates to properly display non-HTML characters in email subject lines.

### DIFF
--- a/brambling/tests/functional/test_invites.py
+++ b/brambling/tests/functional/test_invites.py
@@ -20,9 +20,9 @@ class InviteTestCase(TestCase):
         self.assertEqual(len(mail.outbox), 1)
         self.assertEqual(mail.outbox[0].subject, "{} has invited you to edit {}".format(invite.user.get_full_name(), event.name))
     def test_subject__event_editor_apostrophe(self):
-	event= EventFactory(name="James's Test Event")
-	invite=InviteFactory(content_id=event.pk, user=PersonFactory(given_name="Conan",surname="O'Brien"))
-	self.assertEqual(len(mail.outbox), 0)
+        event= EventFactory(name="James's Test Event")
+        invite=InviteFactory(content_id=event.pk, user=PersonFactory(given_name="Conan",surname="O'Brien"))
+        self.assertEqual(len(mail.outbox), 0)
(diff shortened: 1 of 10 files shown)


brambling/tests/functional/test_invites.py
```diff
@@ -19,7 +19,13 @@ def test_subject__event_editor(self):
         invite.send(Site('test.com', 'test.com'), content=invite.get_content())
         self.assertEqual(len(mail.outbox), 1)
         self.assertEqual(mail.outbox[0].subject, "{} has invited you to edit {}".format(invite.user.get_full_name(), event.name))
-
+    def test_subject__event_editor_apostrophe(self):
+	event= EventFactory(name="James's Test Event")
```
melinath: The indentation here is messed up. Looks like the indentation inside this function (and all the test functions in this PR) is using tabs instead of spaces. One thing that might help is installing the EditorConfig plugin for sublime text – then it'll automatically make sure your code conforms to our style guide.

I'm actually a little confused as to how these tests are passing; usually python complains about mismatched indenting...

brambling/tests/functional/test_mailers.py
```diff
@@ -2,7 +2,7 @@
 
 from zenaida.templatetags.zenaida import format_money
 
-from brambling.mail import OrderAlertMailer
+from brambling.mail import (OrderAlertMailer)
```
melinath: It would be great if you could revert this line for clarity.

brambling/tests/functional/test_invites.py
melinath: There should also always be an empty line between the end of one method and the beginning of the next. It makes reading the code easier.

melinath: For python code, we conform to Django's style guide: https://docs.djangoproject.com/en/1.9/internals/contributing/writing-code/coding-style/

melinath:
This looks great from a functional perspective! Just needs the code style cleaned up & it should be good to go.

melinath:
It looks like the name-rename branch added some tests that use given_name and surname, so they need to be updated for these changes as well. :-)

PR #737, opened 2016-01-09
Updated org percentage to 2.5%

JCattera:
Changed models.py so that Organization.default_application_fee_percent and Event.default_application_fee were now 2.5 instead of 1.5, and also updated about.html and pricing.html in templates/brambling accordingly.


melinath:
This will need a migration as well.

melinath:
Otherwise, looks good!

PR #738, opened 2016-01-11
Removed "apply discount" tab and javascript handling from organizer order detail page.

JCattera:
Also removed related handling and views in views/organizer.py and urls.py.


melinath:
It looks like this is accidentally based off the org-percentage branch. There are two ways this could be fixed:

1. Reset to master and then cherry-pick the last commit

```
git checkout remove-apply-discount
git reset --hard master
git cherry-pick c053e8c
git push --force origin remove-apply-discount
```

2. Do an interactive rebase and remove the extraneous commit
```
git checkout remove-apply-discount
git rebase -i master
# comment out commit 0253086
git push --force origin remove-apply-discount
```

melinath:
It looks like the Remove / Apply discount views for organizers were just 100% vestigial, and that the real meat is in the discount_form context var that OrderDetailView provides.

melinath:
Why did you close this?

JCattera:
i'm really sorry - I didn't mean to close it and i just did the first method you suggested, so i'm surprised that it closed itself.

melinath:
It's not that big a deal. I wouldn't have expected it to close itself either! Very strange.

PR #741, opened 2016-01-15
Changed discount form on shop page so pressing enter would submit the discount.

### LINKED ISSUE #689: "Enter" on discount form doesn't apply discount
With this field highlighted, I would expect pressing enter to submit the form and apply the discount, but you have to click to submit. This probably also has accessibility implications.

screenshot 2015-10-29 13 01 57

JCattera:
Resolves #689 by wrapping the discount field in shop.html with a form tag and having _user_discount_js.html refer to the button's 'submit' action.

### DIFF
--- a/brambling/templates/brambling/event/order/_use_discount_js.html
+++ b/brambling/templates/brambling/event/order/_use_discount_js.html
@@ -2,7 +2,7 @@
     $(function(){
         $('#discountForm').on('submit', function(e){
             e.preventDefault();
-            var discountCodeButton = $(this),
+            var discountCodeButton = $(this).find('button'),
                 discountCode = $('#discountCode'),
                 code = discountCode.val(),
                 url = '/{{ event.organization.slug }}/{{ event.slug }}/order/use-discount/' + code + '/';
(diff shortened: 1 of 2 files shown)


brambling/templates/brambling/event/order/_use_discount_js.html
```diff
@@ -1,6 +1,6 @@
 <script>
     $(function(){
-        $('#discountCodeButton').on('click', function(e){
+        $('#discountForm').on('submit', function(e){
             e.preventDefault();
             var discountCodeButton = $(this),
```
harrislapiroff: This will need to be corrected. In the old version of the code, `this` referred to the `<button>`, but in the new version `this` refers to the `<form>`. The button will have to be identified some other way. (`$(this).find('button')` might work.)

melinath:
@JCattera I didn't notice Harris's comment and I rebased this branch off master... you'll need to run the following commands before continuing work on this branch:

```
git checkout enter-discount
git fetch
git reset --hard origin/enter-discount
```

Issue #743, opened 2016-01-22
Event attendee invites are broken - rely on purged method


melinath:
When Event.create_order was purged, we missed one. If you send an invite to attend an event, and it's accepted, the attendee will get a 500 error.

PR #744, opened 2016-01-22
Event admin split

### LINKED ISSUE #726: Split event settings into multiple pages
Even on desktop (but especially on mobile) the event settings page is long and hard to deal with. It's not really clear what you're gettting when you go there or where on the page you'll be able to find it. I think a good solution could be to split it into several sections:

1. Basic information - name, urls, description, dates, location, currency?, timezone
2. Design - banner image(s) (see #700) and any other design options we add in the future
3. Registration - cart expiration, event privacy, etc.
4. Permissions - editors; this could maybe stay in basic info?
5. Payment - currency?, check postmark date, list of payment options (maybe even the actual form from org for people with the right permissions?)

Side note: 1, 2, and 3 could be part of an initial setup process (see #462).

But the big questions for me are:

1. Is this really a thing we should do? I think it is, especially as we add even more configuration options to events, but I'd love a second.
2. Where the fuck should we put the navigation? We already have 8 items in the admin nav; this would take that up to 12. We could maybe save some space by putting 'copy event' somewhere else - maybe on the basic information page? (It should probably go wherever we end up wanting to put the 'delete' button - see #254.) We could do some kind of nested navigation, maybe tabs for the various new pages, but that feels like it could just make things even more complicated.

Though along those lines, we could make the *entire* nav for events more hierarchical - for example, splitting it up into Summary, Settings, "Related Objects", and Reports. But again, that seems like it makes things just harder to find. :-/

melinath:
Resolves #726. Splits the event settings into a few separate sections, and splits the nav into a few (collapsible) sections. A few of these sections are currently a bit sparse, but will be fleshed out as other tickets that have been waiting on this change drop.

Known issue: Event attendee invites have problems unrelated to this branch (see #743).

This might need some design touchup / field reordering; I'd be interested to hear what people think.


harrislapiroff:
This looks cool. I don't think the Danger Zone section makes much sense. From a user-perspective A. you have no idea what settings are inside it without actually going to it and B. copying an event doesn't seem very dangerous.

harrislapiroff:
Also, collapsing doesn't make much sense if it doesn't persist across page loads.

harrislapiroff:
Publish Event Shop and View Event Shop take up a lot of prominent space considering how infrequently they will be used. Thinking about a better place to put them.

harrislapiroff:
I think keeping the date in the sidebar would also be a good idea. I can imagine (theoretically) an organization having an event with the same name every year (with different slugs). An organizer would want to be able to tell at a glance which year it is of that event.

harrislapiroff:
I think we should move some of these buttons and information to a top bar. But maybe in a different PR.

img_7745

melinath:
@harrislapiroff the publish button is only there until the event is published, so I think the prominence makes sense.

melinath:
Also, yeah, I was having some trouble coming up with good names for the nav sections...

harrislapiroff:
Oh interesting, good point. Yeah. I'm playing with other layout stuff; why don't you take on the collapse stuff.

harrislapiroff:
I decided to do most of the layout stuff in this pr after all. :flushed:

harrislapiroff:
I think I'm done meddling with this. Once the collapse stuff is done, I'm happy with this feature.

melinath:
@harrislapiroff I like the design improvements! If the collapse stuff looks good to you, let's merge!

harrislapiroff:
:+1:

PR #746, opened 2016-01-28
Order create purge

### LINKED ISSUE #743: Event attendee invites are broken - rely on purged method

melinath:
Resolves #743 – thanks, @JCattera

### DIFF
--- a/brambling/tests/functional/test_invites.py
+++ b/brambling/tests/functional/test_invites.py
@@ -157,3 +157,6 @@ class InviteAcceptViewTestCase(TestCase):
         with mock.patch.object(wraps=Order.objects.for_request, target=Order.objects, attribute = 'for_request') as for_request:
             view.handle_invite()
         for_request.assert_called_once_with(create=True, request=view.request, event=view.content)
+        orders = Order.objects.all()
+        self.assertEqual(len(orders), 1)
+        self.assertEqual(orders[0].person, view.request.user)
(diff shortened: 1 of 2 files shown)


brambling/tests/functional/test_invites.py
```diff
@@ -137,3 +145,15 @@ def test_context_data__no_invite(self):
...
+        view = InviteAcceptView()
+        event = EventFactory()
+        view.invite = InviteFactory(content_id=event.pk, kind=Invite.EVENT, user=PersonFactory(first_name="Conan",last_name="O'Brien"))
+        view.content = view.invite.get_content()
+        view.request = RequestFactory().get('/')
+        view.request.user=PersonFactory()
+        self._add_session(view.request)
+        with mock.patch.object(wraps=Order.objects.for_request, target=Order.objects, attribute = 'for_request') as for_request:
+            view.handle_invite()
+        for_request.assert_called_once_with(create=True, request=view.request, event=view.content)
```
melinath: This is great; we also should check the db state as long as we're here - i.e. that the order got created and has the expected properties.

brambling/views/core.py
```diff
@@ -115,7 +115,7 @@ def handle_invite(self):
         invite = self.invite
         content = self.content
         if invite.kind == Invite.EVENT:
-            content.create_order(self.request.user)
+            return Order.objects.for_request(event=content, request=self.request, create=True)
```
melinath: This probably shouldn't return - just so that it matches the other sections of this func.

PR #748, opened 2016-01-29
Corrected tooltip placement in event admin header

melinath:
What it says on the box.


melinath:
^ This got messed up in the redesign for #744 :-(

Issue #751, opened 2016-02-02
Support for merging conflicting orders


melinath:
Right now, conflicting orders prevent users from seeing all of their purchases, which is *basically* a data loss bug. We need to fix this ASAP.

I'd suggest that we provide a merging interface that basically assigns all the transactions & items on the old order to the new order and strips their attendees, then directs the user to the assignment page. Maybe we should also transfer attendees? But I wouldn't want to end up in a situation where the event has duplicate attendees mucking up the works.

We should also transfer hosting data that is on the old one but not the new one. And we should provide a message saying what was transferred. And we should delete the old order.

melinath:
Report here: https://www.facebook.com/events/1558873821063246/permalink/1691611431122817/

harrislapiroff:
Ultimately we should probably support associating multiple email addresses with an account. Maybe doesn't need to be immediately though.

melinath:
That also wouldn't solve this issue... ;-)

harrislapiroff:
Oh. Maybe I misunderstood!

melinath:
Resolved with merge of #753.

PR #753, opened 2016-02-02
Permit merging of conflicting orders

chigby:
Attempts to address #751.

Work in progress. Still needs:
 - UI, or a way to post to this view
 - Redirect back to the proper place (the attendee assignment page?)
 - Merging of `EventHousing` data (this is potentially tricky)
 - Message reporting what happened.


brambling/views/user.py
```diff
@@ -296,11 +297,38 @@ def get_context_data(self, **kwargs):
         context = super(ClaimOrdersView, self).get_context_data(**kwargs)
         context.update({
             'claimable_orders': self.request.user.get_claimable_orders().select_related('event__organization'),
-            'unclaimable_orders': self.request.user.get_unclaimable_orders().select_related('event__organization'),
+            'unclaimable_orders': self.request.user.get_mergeable_orders().select_related('event__organization'),
```
melinath: template key also needs to be renamed

brambling/tests/functional/test_merge_orders.py
```diff
@@ -0,0 +1,103 @@
...
+            response = self.view.post(self.view.request)
+
+    def test_should_be_not_found_if_user_not_confirmed(self):
+        self.view.request = self.factory.post('/', {})
+        self.view.request.user = PersonFactory(confirmed_email='')
+
+        with self.assertRaises(Http404):
+            response = self.view.post(self.view.request)
+
+    def test_should_be_not_found_if_order_claimable_or_does_not_exist(self):
```
melinath: if_order_claimable should probably be a separate test that has a claimable order.

melinath:
I think we can probably not try to transfer any attendee or housing info (for now), as long as we direct people to verify that their registration info is correct. The priority for this ticket is making sure people have access to everything they paid for...

But I keep flip-flopping on this. Maybe it would be good to transfer attendees, and transfer EventHousing if one doesn't already exist?

melinath:
Having to re-enter data would also be pretty frustrating.

brambling/tests/functional/test_merge_orders.py
```diff
@@ -0,0 +1,127 @@
...
+        with self.assertRaises(Http404):
+            response = self.view.post(self.view.request)
+
+    def test_should_be_not_found_if_user_not_confirmed(self):
+        self.view.request.user = PersonFactory(confirmed_email='')
+
+        with self.assertRaises(Http404):
+            response = self.view.post(self.view.request)
+
+    #def
```
melinath: to remove

chigby:
@melinath @harrislapiroff I think this is ready to review -- currently it is not transferring attendees or eventhousing.

melinath:
This is looking good!

A couple things:

1. Could the "Merge" button go on the right like the "Claim" button does?

    screen shot 2016-02-04 at 12 19 38 am

2. Now that merging is a thing, the count in the menu should probably include the mergeable count as well.

    screen shot 2016-02-04 at 12 21 34 am

3. I'm feeling more and more like we should transfer attendees (and hosting info if possible). Otherwise we'll just have another data loss bug down the line where someone does the transfer and doesn't re-enter their attendees, resulting in purchased items that aren't attached to anyone – and potentially attendees who vanish from the database! Event admins will be able to deal much better with occasional duplicates.

    We can still ask people to go through and verify their data; just less danger this way.

chigby:
@melinath by "hosting info" do you mean EventHousing objects? If so I think that might be hard to do because an Order can only have one EventHousing, so it's not clear which one should supercede the other. There's going to be some data lost either way.

melinath:
@chigby Clarification: I think we should transfer it if the one it's being merged into *doesn't* have an EventHousing object attached, and just let it get deleted otherwise. It's data loss, but not *as* bad? If we allowed multiple hostings per order (which we should someday) then I'd want to switch to copying it over. But until then this should be fine.

melinath:
The modal can even give a specific warning if the EventHousing conflicts & will be lost, and otherwise just give a general notice that folks should verify their data.

chigby:
@melinath Okay, I think this is ready to review

melinath:
This looks great. Merging.

PR #759, opened 2016-02-13
Permission improvements

### LINKED ISSUE #747: Read-only permissions for events
We've had a request for more advanced permissions around events - specifically, being able to give certain users the ability to view and download data without being able to edit the event or edit users' orders.

* [x] Remove references to editable_by
* [x] Remove references to Invite.EVENT
* [x] Remove references to Invite.EVENT_EDIT
* [x] Remove references to Invite.EVENT_VIEW
* [x] Remove references to Invite.ORGANIZATION_OWNER
* [x] Remove references to Invite.ORGANIZATION_EDIT
* [x] Remove references to Invite.ORGANIZATION_VIEW
* [x] Remove references to Invite.TRANSFER
* [x] Set up formset js for event & org permissions
* [x] Switch perms cache to be per-request, not global - i.e. write a perms backend to hook into django
* [x] Add tests of perms handling
* [x] Allow removal of Owners, but not of the last owner
* [x] Add templates for new invite types (and tests for them)
* [x] Set up actual usage of the permissions
* [x] Remove old permissions fields
* [x] Fix tests that break after old fields are removed
* [x] Look for other bits that break after old fields are removed - owner, editors, organization__owner,  #organization__editors, additional_editors
* [x] Double-check actual usage of the permissions
* [ ] On deploy: add brambling perms backend to settings.py

Tickets to create:
* [x] Switch views & templates to pass all user perms for the object into the context & use that rather than using organization_editable_by and friends.

melinath:
Resolves #747. Refactored permissions to be handled by an m2m throughs with an additional `role` field that has three options for org (owner, editor, view) and two options for event (editor, view). Switched from the increasingly cumbersome method of checking against a list of roles to having a object-level perms backend that returns a list of up to three permission types `('change_permissions', 'edit', 'view')` for the given user. This is cached on the user object, like Django's built-in permissions, which makes it efficient and thread-safe. Doing things this way will also be much more extensible in the future, since we can just add new permission types and assign them to whatever roles we want (instead of having to write a new method every time we introduce a new thing.)

Also refactored invites to be more modular, so that it will (hopefully) be clearer what code belongs together. It was getting a little out of hand with all the different invite types...


brambling/utils/invites.py
```diff
@@ -0,0 +1,510 @@
...
+        return reverse('brambling_event_shop', kwargs={
+            'event_slug': content.slug,
+            'organization_slug': content.organization.slug
+        })
+
+    def manage_allowed(self):
+        # Invites can be created or modified by people who
+        # can edit the event.
+        event = self.get_content()
+        return self.request.user.has_perm('edit', event)
```
chigby: I sort of worry our permissions (i.e. what users can and cannot do) aren't distinct enough from the roles (i.e. what users are). The system you have made decouples these two ideas (or seems to, anyway). This function is ripe for something like `has_perm('manage_invites', event)` -- is there a reason we shouldn't have that?

melinath: That's exactly the direction I think we should go! But it would have made this PR even more massive, and it'll be easy to implement things like that later on.

brambling/models.py
```diff
@@ -482,17 +551,44 @@ def get_absolute_url(self):
     def get_liability_waiver(self):
         return self.liability_waiver.format(event=self.name, organization=self.organization.name)
 
-    def editable_by(self, user):
-        return (
-            self.organization.editable_by(user) or (
-                user.is_authenticated() and
-                user.is_active and
-                self.additional_editors.filter(pk=user.pk).exists()
+    def get_permissions(self, person):
```
chigby: This function feels slightly unwieldy, with the potential to become more unwieldy if we ever add more permissions, more roles, or more membership types (or need to change some of the above). I'm not sure I have any super-good ideas about what to do instead, which is frustrating because all that's really going on here is taking, like, the union of { OrganizationMemberPermissions, EventMemberPermissions, SuperuserPermissions }.

I might need to ponder this more, but I think it's worth pondering. I don't want us to get into a situation where it's hard to tell who has permission to do what, which I think this might be on the verge of.

brambling/auth_backends.py
```diff
@@ -0,0 +1,41 @@
+from brambling.models import (
+    Organization,
+    OrganizationMember,
+    Event,
+    EventMember,
+)
```
chigby: Unused imports.

brambling/models.py
```diff
@@ -80,6 +79,10 @@
 UNAMBIGUOUS_CHARS = 'abcdefghijkmnpqrstuvwxyzABCDEFGHJKLMNPQRSTUVWXYZ23456789'
 
 
+class CACHED_DOES_NOT_EXIST:
+    pass
```
chigby: Not sure what this does.

brambling/models.py
```diff
@@ -299,6 +302,30 @@ def create_defaults(app_config, **kwargs):
             ])
 
 
+class OrganizationMember(models.Model):
+    EDIT = '1-edit'
+    VIEW = '2-view'
+    OWNER = '0-owner'
```
chigby: This may just be a convention I'm not aware of, but what are these numbers, here?

chigby:
The dropdown gets cut off for me if there aren't enough people in the list:
screen shot 2016-02-18 at 16 20 23

Happens in both firefox and chrome.

chigby:
I also wonder if eventually we'd want to make it possible to add permissions to existing users without sending an invitation. Like, right now there aren't that many users to begin with so most of the time an email would be sent, but if that email address is within Dancerfly already, seems like clicking the link is extra work?

chigby:
@melinath This is probably all the comments I have.

brambling/models.py
melinath: We could make this function more wieldy (maybe) by adding a bunch of permissions checkboxes to the models... but that would add its own complications. I think this is as good as we can do for now? Definitely way better than the old version.

brambling/models.py
melinath: That is a leftover from an intermediate stage. :-/

brambling/models.py
melinath: They allow ordering by the role field to return a list in order of descending permissions.

melinath:
@harrislapiroff @nmorduch do you have any advice on the select cutoff issue @chigby mentioned?

brambling/models.py
chigby: I do think it's fine for now.

chigby:
I wonder if it's a `z-index` problem?

brambling/models.py
chigby: Ah, okay, cool. Though this makes me think of another thing.

melinath:
It's bound to be an overflow:hidden problem. I'm just not sure if I should turn that off vs. looking for some other solution...

brambling/models.py
```diff
@@ -378,14 +421,32 @@ def stripe_live_can_connect(self):
...
+class EventMember(models.Model):
+    EDIT = '1-edit'
+    VIEW = '2-view'
+    ROLE_CHOICES = (
+        (EDIT, 'Can edit event'),
+        (VIEW, 'Can view event'),
+    )
+    event = models.ForeignKey('Event')
+    person = models.ForeignKey('Person')
+    role = models.CharField(max_length=6, choices=ROLE_CHOICES)
```
chigby: How strongly do you feel about using CharField here? My attitude is, if we're using postgres, to just put store strings in the `text` type and not have to worry about length enforcement. Especially considering this field doesn't accept user data, limiting ourselves to 6 characters feels stingy if we ever want to add more roles.

melinath: I didn't realize that [Postgres performance for varchar & text was the same](http://www.depesz.com/2010/03/02/charx-vs-varcharx-vs-varchar-vs-text/) – that's neat! I think I'd still like to stick with a varchar, though. As much as possible, I would prefer to not tie ourselves to a particular database. I also think that CharField is more representative of what we're doing here, esp. for new general Django developers. (Really, if we were going to switch to a different field, I'd want us to investigate getting a SET implementation.)

I don't think that the limit will be an issue, since migrations are so easy. If we need to lengthen it we can. And I'm not sure that Django's TextField supports choices.

chigby: Works for me.

melinath:
Okay, so it's not overflow: hidden, it's overflow: scroll. But I can't find it! And Chosen doesn't support other more magical solutions... definitely need some help here.

harrislapiroff:
Putting a guess here before actually looking at anything: it has to do with it been a bootstrap responsive table.

harrislapiroff:
Yeah, it's because of the `table-responsive` class. The weird thing is that `table-responsive` only has `overflow-x: auto` and doesn't dictate anything about the Y direction. So I'm surprised that it's creating a vertical scroll as a result, but it definitely gets fixed when I remove that declaration.

If I don't find a better solution, the solution will be to remove the `table-responsive` class, because it's not really like the site (and particularly the organizer interface) is particularly responsive anyway.

harrislapiroff:
(It would be nice if chosen let you choose the parent element of the dropdown, the way that bootstrap does for tooltips and popovers...)

harrislapiroff:
Ah, from w3c spec:

> The computed values of ‘overflow-x’ and ‘overflow-y’ are the same as their specified values, except that some combinations with ‘visible’ are not possible: if one is specified as ‘visible’ and the other is ‘scroll’ or ‘auto’, then ‘visible’ is set to ‘auto’.

I guess that makes some sense.

harrislapiroff:
Okay, fixed. Bouncing back to @chigby.

chigby:
@harrislapiroff is the dropdown working for you? It's not for me. 
screen shot 2016-02-23 at 16 59 32

harrislapiroff:
Let me look again.

harrislapiroff:
Whoopsie. All done now, @chigby.

(I wonder if we could combine those two tables into a single include template?)

chigby:
@harrislapiroff Okay, working for me everywhere now. I'd say this is ready to merge? If you want to merge the templates, though, go for it.

chigby:
And by "go for it" I mean: let me know if you want to and I'll hold off on merging it.

harrislapiroff:
Go for it! :D

PR #773, opened 2016-02-26
Flake8 in tests

melinath:
Adds flake8 run to tests. Currently just ignores lines that are too long, since we don't enforce that strictly. All things considered, I think we've done pretty well :-) This will ensure our continued success!


chigby:
This looks good to me! Great to have no PEP8 warnings except line lengths.

PR #774, opened 2016-02-26
Fix per-transaction receipts in admin

### LINKED ISSUE #761: Support sending per-transaction receipts from the admin interface
In the switch to per-transaction receipts, it looks like we missed updating the organizer UI; it still just has one button for the entire order. This causes a 500 error if you try to send, since we pass summary_data and don't pass a transaction. We need one button for each transaction (in the actions area I guess).

chigby:
Fixes #761.

This pull request makes `SendReceiptView` send an email for transactions instead of orders, which is how `OrderReceiptMailer` works now. 

The button to send the receipt shows up once per transaction on the order details page, like this: 

screenshot from 2016-02-26 18-06-30

There may be a better design for this, but it seems okay to me.


chigby:
@harrislapiroff @melinath This is ready for review.

melinath:
I haven't tested yet, but this all looks good. Just one thing: We should restrict display of this button & access to the view to payment transactions – i.e. this email is not intended for refunds.

chigby:
@melinath Should be good to test, we're now only showing the button for payments.

melinath:
I made a couple minor tweaks. I'm down with merging this PR; if my changes look good to you, go ahead!

PR #778, opened 2016-02-28
Django 1.8

melinath:
Upgrades us to django 1.8!


melinath:
Okay actually done now.

chigby:
Well, the tests are passing and a brief click around of the various parts of the site didn't crop up any errors. Upgrade seems pretty straightforward, which is good!

I noticed the latest Django 1.8 release is 1.8.11 but we're using 1.8.9 here. Do you think it's worth going to the latest? Seems to just fix some bugs: 
 - https://docs.djangoproject.com/en/1.8/releases/1.8.10/
 - https://docs.djangoproject.com/en/1.8/releases/1.8.11/

PR #780, opened 2016-03-02
Prevent charging stripe if amount is negative.

chigby:
Fixes error we saw on production site: `'NoneType' object has no attribute 'card'`

Context:
`/var/www/env/src/django-brambling/brambling/forms/orders.py in _post_clean`

```
self <brambling.forms.orders.OneTimePaymentForm object at 0x7fa7bf6e2990>

kwargs
{'amount': Decimal('-94.00'),
 'event': <Event: {{name removed}}>,
 'order': <Order: Order object>}
```

Other notes:

 * If the amount is negative, we can't charge and the form code
   assumes that you can charge.
 * Setting `self._charge` to `None` by default to mimic the Dwolla
   form functionality.  Not totally sure if needed.

### DIFF
--- a/brambling/forms/orders.py
+++ b/brambling/forms/orders.py
@@ -493,19 +493,17 @@ class DwollaPaymentForm(BasePaymentForm):
 
     def _post_clean(self):
         if 'dwolla_pin' in self.cleaned_data and 'source' in self.cleaned_data:
-            self._charge = None
-            if self.amount > 0:
-                try:
-                    self._charge = dwolla_charge(
-                        account=self.dwolla_account,
-                        amount=float(self.amount),
-                        order=self.order,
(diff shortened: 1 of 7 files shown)


melinath:
This needs to be more robust & needs tests. This only changes OneTimePaymentForm, but the same issue would also affect SavedCardPaymentForm. But also, if I'm reading the code right, defaulting `self._charge` to `None` just means that if the form is "valid" – which means it'll save, which will pass `self._charge` to `Transaction.from_dwolla_charge` or `Transaction.from_stripe_charge`, which don't know how to handle `None`.

I think the solution is to just add an error to each of the three forms if the amount <= 0.

chigby:
@melinath What do you think of this? Ready for review.

PR #783, opened 2016-03-09
Make Order forms always display, regardless of "collect survey data"

chigby:
Fixes issue raised in feedback.

The collect survey data checkbox now explicitly controls only the
built-in questions (how did you hear about it, distribution
promotional materials).  If there is an order form, there will always
be a survey step with that form on it.


brambling/views/orders.py
```diff
@@ -151,7 +151,8 @@ class SurveyStep(OrderStep):
 
     @classmethod
     def include_in(cls, workflow):
-        return workflow.event.collect_survey_data
+        return (workflow.event.collect_survey_data or
+                workflow.event.forms.filter(form_type=CustomForm.ORDER).exists())
```
melinath: There's a small chance that this should check for fields rather than a form, but... w/e. If we get there it'll be a pretty drastic overhaul of the custom form stuff anyway, so no use trying to predict what we'll want to do.

PR #786, opened 2016-03-14
Manage event dropdown

### LINKED ISSUE #327: Dropdown for "manage event" button
that would give links to all the pieces of manage event because you're generally trying to get somewhere other than the summary and it's annoying

chigby:
This pull request adds: a super-basic dropdown with shortcut links to the different event management areas. Seems like it gets the job done, but could potentially use some bolt-tightening if there's a better way to structure the markup, responsiveize it, give it some visual flair, and so on.

Resolves #327.

### DIFF
--- a/brambling/filters.py
+++ b/brambling/filters.py
@@ -124,5 +124,7 @@ class OrderFilterSet(FloppyFilterSet):
             ('code', 'Code'),
             ('-code', 'Code (descending)'),
         ]
+
+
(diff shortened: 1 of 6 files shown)


nmorduch:
This is good! I separated the dropdown into its own include so that I could also stick it on the shop page :D

nmorduch:
Hm hm hm I moved the manage button into that image on the dashboard but I suppose it's unhappy about having a link within a link

nmorduch:
BOOM.

Issue #244, opened 2014-10-27
Allow organizers to email addresses


harrislapiroff:
We should allow organizers to send email to attendees, either by:
- providing our own email interface: might be nice because we can append our own "You're receiving this email because ____. If you did not register for this event, go here." or something and because we can ensure that organizers don't run afoul of CAN-SPAM and don't email 200 people without BCCing
- providing a copy/paste list of email addresses: easy, we don't have to implement much or pay a transactional mail service based on our users' emailing sending habits

melinath:
For now, it would be easiest to provide a copy/paste list of email addresses. Actually, they could already go to the attendee list and download a CSV of just email addresses...

melinath:
From @nmorduch:

>This could be a button that copies an email list to your clipboard (too easy to forget to bcc)
or, better, it could be a way to send plain-text emails through us (also a good way to remind people we exist!)
>
>OR we could have a mailto link that automatically has everyone in the bcc field??????

melinath:
Given that people can already copy-paste email addresses from the CSV and that adding anything to the UI is going to be a complex issue, I think we should probably just go ahead and implement an email sending interface (backed by usability research, of course.)

melinath:
(It would be cool also to be able to route incoming emails to support+<org/event slugs>+<order code>@dancerfly.com into an inbox, or just forward straight to organizers.)

PR #788, opened 2016-03-14
Link to Email Attendees

chigby:
See #244 for motivation.

New features: a Communication panel on the event summary page that has some explanatory text and a button `mailto` link with all of the attendee emails in the BCC field.

screen shot 2016-03-14 at 15 24 41

Discussion: I like the idea of a dedicated area for communication tools, potentially this could be expanded later to include actual email sending functions (as noted in #244) but right now just a link seems fine. This could also be on its own "tab" with a link on the side menu, but I thought that might be too much for just a single link. This, as well as other potential "homes" for the link are up for debate.

Also the documentation could use a once-over from someone.

### DIFF
--- a/brambling/templates/brambling/event/organizer/summary.html
+++ b/brambling/templates/brambling/event/organizer/summary.html
@@ -157,6 +157,18 @@
 					</a>
 				</div>
 			</div>
+			<div class='panel panel-default'>
+				<div class='panel-heading'>
+					<h4 class='panel-title'>
+						Communication
+					</h4>
+				</div>
+				<div class="panel-body">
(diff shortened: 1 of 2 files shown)


melinath:
Just had an event ask about this... I think I'd prefer to have a way for people to copy the email addresses? I'm not sure we can count on a mailto: link working the way we expect - slash it is fairly prescriptive.

Still, in the absence of any kind of internal mail sending tool... although maybe we _should_ just provide an internal mail-sending tool...

melinath:
Closing this for now. We should do more research slash potentially implement an email sending interface instead.

PR #795, opened 2016-03-28
Implement staging and production tiers for fabric.

### LINKED ISSUE #790: Set up tasks to be per-server rather than using -H on the command line
Right now, we deploy like so:

```
fab -H <server_name> deploy # Deploys a clone of production to the given server
```

Which is fine, except that it doesn't let us do things like run a task as if the current server is a staging vs production server. How this _should_ work:

```
fab production deploy # Deploys a clone of production to the production server (which fab knows something about)
fab staging deploy # Deploys a staging version to the staging server (which fab knows something about0
```

See the following for information / inspiration:

* http://docs.fabfile.org/en/latest/usage/execution.html
* http://djangotricks.blogspot.com/2010/05/deployment-using-fabric.html

chigby:
Adds two tasks for switching between the different deployment tiers we
currently support: `staging` and `production`.  These tasks set values
on `env` according to that particular environment.  At this current
moment it's just the host to run the commands on, but it could be
anything.

This code now requires a tier to be set before running any tasks.  So
you can no longer run tasks with `fab -H <hostname> deploy`, you must
run, e.g., `fab staging deploy`.

Resolves #790.

### DIFF
--- a/fabfile.py
+++ b/fabfile.py
@@ -2,7 +2,7 @@ import os
 
 from fabric.api import task, run, sudo, cd, env
 from fabric.contrib.files import exists
-from fabric.operations import local, put
+from fabric.operations import local, put, require
 
 
 env.use_ssh_config = True
@@ -13,13 +13,41 @@ DEFAULT_BRANCH = 'master'
 CURRENT_DIR = os.path.dirname(__file__)
 
 
+TIERS = {
+    'production': {
+        'hosts': ['root@104.131.92.59'],
+    },
+    'staging': {
+        'hosts': ['root@162.243.226.226'],
+    },
+}
+
+
+def tier_set(tier_name='staging'):
+    env.tier = tier_name
+    for option, value in TIERS[env.tier].items():
+        setattr(env, option, value)
+
+
+@task
+def production():
+    tier_set('production')
+
+
+@task
+def staging():
+    tier_set('staging')
+
+
 @task
 def install_salt():
+    require('tier', provided_by=(staging,production))
     run("curl -L https://bootstrap.saltstack.com | sudo sh -s -- -P git a6c0907")
 
 
 @task
 def sync_pillar(branch_or_commit='master'):
+    require('tier', provided_by=(staging,production))
     if not os.path.exists(os.path.join(CURRENT_DIR, 'pillar')):
         local("git clone {} pillar".format(PILLAR_REPO_URL))
     if not exists('/root/.ssh/id_rsa'):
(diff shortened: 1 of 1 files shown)


chigby:
Potentially there could also be a "local" tier, though I'm not sure running tasks locally is something people use fabric for in this case.

fabfile.py
```diff
@@ -13,13 +13,41 @@
...
+
+
+@task
+def staging():
+    tier_set('staging')
+
+
 @task
 def install_salt():
+    require('tier', provided_by=(staging,production))
```
melinath: This line (and similar) should have a space after the comma in `staging,production`.

Also, this... doesn't seem to do anything? i.e. if I remove 'staging' from the list for deploy, I can still run `fab staging deploy`...

fabfile.py
```diff
@@ -13,13 +13,41 @@
 CURRENT_DIR = os.path.dirname(__file__)
 
 
+TIERS = {
+    'production': {
+        'hosts': ['root@104.131.92.59'],
```
melinath: This is weird to me, but I guess ip addresses are not really secret. Though I do still feel pretty odd about the fact that we include the username here.

melinath: I don't think there would be a way around that besides relying on people having the correct ssh config set up. Which removes it from version control, which is not good.

chigby: I, too, feel a little weird about having the username there, but I don't _think_ it's exposing us to much, if any, harm?

fabfile.py
chigby: D'oh, shoulda run flake8.

What this line does is require that the tier variable be set. The `provided_by` flag I think is just to tell the user what they can do to set it if it's not, e.g. this help message:

```
$ fab -H some.website.com deploy
[some.website.com] Executing task 'deploy'

Fatal error: The command 'deploy' failed because the following required environment variable was not defined:
    tier

Try running one of the following commands prior to this one, to fix the problem:
    staging
    production

Aborting.
```

melinath: Oh, okay. That makes sense. But also it would be good to have a mechanism to do the other thing i.e. have certain tasks that are forbidden on production.

chigby: It doesn't look to me like fabric has a built-in way to do that, but since "it's just python" we can do something like this, which doesn't seem too bad to me:

```python
@task
def foo():
    if env.tier == 'production':
        abort('cannot be run on production')
    run("uname -a")
```

melinath: Maybe a decorator? @forbid_in_tiers('production')

Issue #775, opened 2016-02-27
Add HSTS headers.


melinath:
We already require people to use https; we should be good web citizens and say so in our headers.

https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security

melinath:
Additional security measures: https://www.ssllabs.com/ssltest/analyze.html?d=dancerfly.com&s=104.131.92.59&latest

melinath:
It looks like (at least some of) these features could be enabled by using the [SecurityMiddleware introduced in Django 1.8](https://docs.djangoproject.com/en/1.9/ref/middleware/#django.middleware.security.SecurityMiddleware)

melinath:
See also: [check --deploy](https://docs.djangoproject.com/en/1.9/ref/django-admin/#cmdoption-check--deploy) management command, which offers suggestions re: production security.

melinath:
Resolved by merge of #796.

PR #796, opened 2016-03-28
Add Strict-Transport-Security header to nginx config.

chigby:
Suggested in #775.

### DIFF
--- a/salt/webserver/nginx.conf
+++ b/salt/webserver/nginx.conf
@@ -25,6 +25,7 @@ server {
     ssl_certificate_key  {{ pillar['files']['crt_dir'] }}dancerfly.key;
     ssl_protocols        TLSv1 TLSv1.1 TLSv1.2;
     ssl_ciphers          HIGH:!aNULL:!MD5;
+    add_header Strict-Transport-Security 'max-age=31536000; includeSubdomains; preload';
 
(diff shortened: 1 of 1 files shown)


melinath:
@chigby is there an advantage to doing things this way rather than using SecurityMiddleware (which also adds some other measures at the same time)?

melinath:
(I also want to make sure that if we do go with this PR we don't lose sight of the other concerns brought up in #775 and not addressed here. I'm not saying they necessarily _should_ be addressed here. Just don't want to forget them.)

chigby:
@melinath From the SecurityMiddleware docs:

> If your deployment situation allows, it’s usually a good idea to have your front-end Web server perform the functionality provided by the SecurityMiddleware. That way, if there are requests that aren’t served by Django (such as static media or user-uploaded files), they will have the same protections as requests to your Django application.

melinath:
@chigby ah, yes. okay. We should be sure then to also add the other cases that SecurityMiddleware would normally handle.

chigby:
Yes, this is a good plan. I'm working on another branch/future PR with the other security improvements from #796, the security middleware, SSL site tester results, and other sources.

melinath:
Sweet, look forward to that branch :-)

PR #797, opened 2016-03-29
Improve SSL and site security

chigby:
This pull request improves SSL handling and other site security features through HTTP headers. The rationale behind most of the changes is explained in the commit messages. Most of them are taken from Mozilla's incredibly useful [SSL Config Generator](https://wiki.mozilla.org/Security/Server_Side_TLS) and the features of Django's [SecurityMiddleware](https://docs.djangoproject.com/en/1.9/ref/middleware/#module-django.middleware.security).

**Note**: these changes to the nginx configuration should be thoroughly tested before being deployed live, as they could potentially cause outages if something is wrong.

Motivating issue: #796


melinath:
It looks like the init.sls for webserver needs to be updated to create the file at `{{ pillar['files']['crt_dir'] }}dhparam.pem`. Probably similar to how the certs are set up. And the pillar data will need to be updated to include the .pem contents.

melinath:
Similarly, this means that we need to split up the SSL cert data from the verification chain in pillar and render it out to two separate files.

melinath:
This looks awesome, looking forward to merging.

melinath:
@emilyhorsman was suggesting we should maybe do a 4096 dhparam key (but that may not be compatible with some older browsers?)

chigby:
A 4096 bit DH parameter size might be a case of "too much security" where we end up using many more CPU cycles for little benefit. Though I suppose that depends on how secure you want to be.

melinath:
Okay, got this working! Seems good, works in all browsers except IE6 and Vista IE8 o_O

https://www.ssllabs.com/ssltest/analyze.html?d=staging.dancerfly.com

PR #802, opened 2016-05-05
Set default ordering for Orders table to 'most recent'

### LINKED ISSUE #801: Sort by completed date by default
See #559. Currently the Order table defaults to sorting by code, which is not what organizers generally want. We should switch the default to completed date.

melinath:
Based on feedback from organizers. Resolved #801.

### DIFF
--- a/brambling/filters.py
+++ b/brambling/filters.py
@@ -119,6 +119,10 @@ class OrderFilterSet(FloppyFilterSet):
         model = Order
         fields = ['providing_housing', 'send_flyers']
         form = forms.Form
-        order_by = ['code', '-code', '-completed_date']
+        order_by = [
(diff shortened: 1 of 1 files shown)


chigby:
I think this is a great change. Tangentially related point, though: sorting by code seems to be ASCII-betical rather than purely alphabetical -- I feel like this might confuse people. Is it possible to sort by, say, `code.upper()`?

melinath:
@chigby that's a good idea, but it's non-trivial :-( See: http://stackoverflow.com/questions/31862420/do-case-insensitive-ordering-with-django-filter

Could we open that as a separate ticket?

chigby:
Sure. I'll merge this.

PR #803, opened 2016-05-09
Implement partial refunds

harrislapiroff:


harrislapiroff:
This branch is not ready. A few things that I know need to be resolved:

- [ ] Tests!!!!
- [ ] Disallow refunds with no items for $0
- [ ] Javascript should not allow dollar amounts that are higher than the transaction's amount refundable
- [ ] Instances of `Transaction.can_refund` should be replaced with `Transaction.get_refundable_amount` or `Transaction.can_refund` should be rewritten to use `Transaction.get_refundable_amount` in combination with checking for unrefunded items on the transaction

brambling/templates/brambling/forms/rows/no_label_no_errors.html
```diff
@@ -0,0 +1,7 @@
+{% extends "brambling/forms/rows/bootstrap.html" %}
+
+{# Blank the label block: #}
+{% block label %}{% endblock %}
+
+{# For the love of god, please render errors *somewhere* though #}
```
melinath: lol

brambling/templatetags/brambling.py
```diff
@@ -0,0 +1,9 @@
+from django import template
+from django.utils.encoding import force_text
+
+register = template.Library()
+
+
+@register.filter
+def to_string(val):
```
melinath: Oh, I forgot this wasn't a builtin. Django does have the |stringformat filter, which could be used to the same effect (i.e. converting an int to a string) – might be better to do that rather than writing a new filter just for this purpose.

brambling/views/organizer.py
```diff
@@ -1182,29 +1186,75 @@ def get_queryset(self):
...
+
+    def form_valid(self, form):
+        transaction = self.transaction
+        # Must be an empty queryset if no items to refund.
+        # If transaction.refund receives `None` it will refund all items.
+        bought_items = form.cleaned_data.get('items', BoughtItem.objects.none())
+
+        # Compare to None to allow for an empty queryset
+        if bought_items is not None:
+            bought_items = BoughtItem.objects.filter(pk__in=[x.pk for x in bought_items])
```
melinath: BoughtItem.objects.none() returns an empty queryset, not None... maybe this is vestigial code?

brambling/views/organizer.py
```diff
@@ -1182,29 +1186,75 @@ def get_queryset(self):
...
+        }
         try:
-            txn.refund(issuer=request.user,
-                       dwolla_pin=request.POST.get('dwolla_pin'))
+            successful_refund = transaction.refund(**refund_data)
         except Exception as e:
-            messages.error(request, e.message)
+            messages.error(self.request, e.message)
+
+        if successful_refund is not None:
```
melinath: If there's an exception, successful_refund will be unset, not None; this will raise a NameError. This should be in an else: clause from the try/except. Can transaction.refund return None?

brambling/views/organizer.py
```diff
@@ -1182,29 +1186,75 @@ def get_queryset(self):
...
+
+            success_message = "Refunded "
+            if refunded_item_count:
+                success_message += "{0} item{1}".format(
+                        refunded_item_count,
+                        pluralize(refunded_item_count))
+            if refunded_item_count and refunded_amount:
+                success_message += " and "
+            if refunded_amount:
+                success_message += format_money(successful_refund.amount.copy_abs(), successful_refund.order.event.currency)
```
melinath: I did not know about copy_abs(). neat!

brambling/views/organizer.py
```diff
@@ -1249,10 +1299,12 @@ def get_forms(self):
                         form.data = self.request.POST
                         form.is_bound = True
                         break
+        self.transaction_forms = [TransactionRefundForm(t)
+                                  for t in self.order.transactions.all()]
         if show_payment_form:
```
melinath: show_payment_form is a marker of whether the user has permission to edit the event. So we should probably rename the var and only display the transaction forms if it's True as well.

brambling/views/organizer.py
harrislapiroff: The `BoughtItem.objects.none()` bit can actually be deleted. This is still necessary because form data returns a list of instances instead of a queryset. (`bought_items.update()` is called in the `.refund()` method)

brambling/views/organizer.py
harrislapiroff: ab-workout

brambling/views/organizer.py
harrislapiroff: Oops. Good point.

brambling/models.py
```diff
@@ -1342,13 +1342,16 @@ def from_dwolla_refund(cls, refund, related_transaction, **kwargs):
     def is_unconfirmed_check(self):
         return self.method == Transaction.CHECK and not self.is_confirmed
 
-    def can_refund(self):
+    def get_returnable_items(self):
+        return self.bought_items.filter(status=BoughtItem.BOUGHT)
+
+    def get_refundable_amount(self):
```
melinath: This should also be called in transaction.refund()

brambling/models.py
```diff
@@ -1342,13 +1342,16 @@ def from_dwolla_refund(cls, refund, related_transaction, **kwargs):
     def is_unconfirmed_check(self):
         return self.method == Transaction.CHECK and not self.is_confirmed
 
-    def can_refund(self):
+    def get_returnable_items(self):
```
melinath: This should also be called in transaction.refund()

brambling/models.py
```diff
@@ -1342,13 +1342,16 @@ def from_dwolla_refund(cls, refund, related_transaction, **kwargs):
...
+    def get_refundable_amount(self):
         refunded = self.related_transaction_set.filter(
             transaction_type=Transaction.REFUND
         ).aggregate(refunded=Sum('amount'))['refunded']
         # None means there are no refunds, which is relevant
         # for 0-amount transactions.
-        return refunded is None or self.amount + refunded > 0
+        return self.amount if refunded is None else self.amount + refunded
 
     def refund(self, amount=None, bought_items=None, issuer=None, dwolla_pin=None):
```
melinath: The beginning of refund should probably be reworked in a couple ways:

- use self.get_refundable_amount()
- Don't return None if there isn't a refundable amount (is there anything out there relying on this behavior?)
- use self.get_refundable_items() for default bought_items.
- default amount to refundable amount if not provided.

brambling/views/organizer.py
melinath: bought_items can't be None, because then all the refundable items on the transaction would be refunded by .refund(). (Though maybe .refund() should be reworked so that it doesn't have that default.)

harrislapiroff: Oh right, good point. It could be an empty list, but otherwise... yeah, both these lines may still be necessary.

harrislapiroff: I guess the `if` isn't.

brambling/views/organizer.py
emilyhorsman: dying 😂

brambling/views/organizer.py
harrislapiroff: Uhhhhhh so we're not actually hiding _any_ of the forms. The `forms` var here gets returned in this method, but it doesn't actually get used anywhere. The forms are stored on the view instance and are _always_ included in template context. But the template checks for edit permissions before displaying the forms. I'll take the same approach to transaction refund forms now, but I wonder if they should never be passed to the context at all if there aren't permissions...

harrislapiroff:
@melinath or @chigby I might need some help writing tests for this

chigby:
I'd pair with you on it, if you want.

brambling/models.py
harrislapiroff: Done

harrislapiroff:
@melinath Okay! @chigby and I tested the hell out of this functionality. It's ready for your review!

brambling/models.py
```diff
@@ -1383,25 +1383,28 @@ def refund(self, amount=None, bought_items=None, issuer=None, dwolla_pin=None):
...
+                    order=self.order,
+                    event=self.order.event,
+                    payment_id=self.remote_id,
+                    amount=amount,
+                    pin=dwolla_pin,
+                )
+                txn = Transaction.from_dwolla_refund(refund, **refund_kwargs)
+
+        # If no payment processor was involved, just make a transaction
+        if amount <= 0 or self.method not in (Transaction.STRIPE, Transaction.DWOLLA):
```
melinath: Wait, why are we allowing amount < 0?

brambling/views/organizer.py
melinath: ping on this. I'd suggest maybe the following:

```
if 'items' in form.cleaned_data:
    bought_items = BoughtItem.objects.filter(pk__in=[x.pk for x in form.cleaned_data['items']])
else:
    bought_items = BoughtItem.objects.none()
```

...not that it probably matters super much in the grand scheme, but it'll be clearer.

brambling/views/organizer.py
```diff
@@ -1249,8 +1298,10 @@ def get_forms(self):
                         form.data = self.request.POST
                         form.is_bound = True
                         break
-        if show_payment_form:
-            forms = [self.payment_form, self.notes_form]
+        self.transaction_refund_forms = [TransactionRefundForm(t)
```
melinath: It looks like there's no provision here for TransactionRefundForm to be instantiated with `request.POST` data, i.e. for displaying validation errors. Just wanted to double-check that this is intentional.

melinath: It looks like there's no provision here for TransactionRefundForm to be instantiated with `request.POST` data, i.e. for displaying previously-submitted values. Just wanted to double-check that this is intentional.

melinath: oh, of course, you don't need that because it uses an ad hoc api endpoint. nvm

brambling/templates/brambling/event/organizer/_refund_form_inner.html
```diff
@@ -0,0 +1,71 @@
...
+						{% endfor %}
+					</div>
+				{% endif %}
+				{% for item in bought_items %}
+					<div class="list-group-item {% if item.status = 'refunded' %}disabled tipped{% endif %}" {% if item.status = 'refunded' %}title="Refunded" data-placement="left" data-container="body"{% endif %}>
+						<label for="id_refund_{{ item.pk }}" class="row label-block">
+							<div class="col-xs-1 text-center">
+								<input
+									type="checkbox"
+									{% if item.pk|stringformat:"s" in form.items.value or item.status = 'refunded' %}
```
melinath: So. Fun fact. If you've submitted a value, form.items.value might be a list of strings (which I believe is why this was added) but on initial page load form.items.value is a list of integers. :-D

melinath: Easiest solution is probably to just check both for now. But this would be a good candidate for a thing that floppyforms should fix...

brambling/views/organizer.py
harrislapiroff: Resolved.

brambling/models.py
harrislapiroff: Resolved.

brambling/models.py
harrislapiroff: Resolved.

brambling/templates/brambling/event/organizer/_refund_form_inner.html
harrislapiroff: Can I just cast both `item.pk` and `form.items.value` to string with `|stringformat:"s"` or will that cause trouble?

brambling/models.py
harrislapiroff: Uhhhhhhhhhhhh. Good question.

harrislapiroff: Fixed.

brambling/templates/brambling/event/organizer/_refund_form_inner.html
melinath: form.items.value is a list; casting it to a string would not be a good idea. For example, then checking for `1` would return true for the list `"[100]"`. Probably better to just add `item.pk in form.items.value` as well.

harrislapiroff: ohhhhh right, I forgot that was a list :\

harrislapiroff:
OK @melinath! It's ready again with all of the above resolved.

PR #815, opened 2016-06-24
Stripe refund webhook

### LINKED ISSUE #380: Add support for stripe refund via webhook
Refunds of stripe charges can happen directly from their website. We should get notification via webhook and synchronize our data.

chigby:
Initial submission, see commit messages for detailed notes.

Resolves #380.

Link to Asana task: https://app.asana.com/0/90578496402390/119551005130791


chigby:
@melinath This is in a situation where it can be reviewed / commented on.

brambling/tests/functional/test_webhooks.py
```diff
@@ -138,3 +281,183 @@ def test_confirmation(self):
...
+
+
+class StripeWebhookBadRequestTestCase(TestCase):
+    def setUp(self):
+        stripe_prep(TEST)
+        self.factory = RequestFactory()
+        self.view = StripeWebhookView.as_view()
+
+    def test_bad_content_type(self):
+        request = self.factory.post('/', content_type='text/xml', data='12345')
```
melinath: Does this data count as valid json?

melinath: Yes.

brambling/views/payment.py
```diff
@@ -120,3 +124,89 @@ def post(self, request, *args, **kwargs):
...
+
+        try:
+            event = stripe.Event.retrieve(stripe_event_id)
+        except stripe.error.InvalidRequestError:
+            raise Http404('Event not found on stripe')
+
+        if event.type != 'charge.refunded':
+            return HttpResponse(status=200)
+
+        with transaction.atomic():
```
melinath: This following section should probably just use [get_or_create](https://docs.djangoproject.com/en/1.9/ref/models/querysets/#get-or-create)

brambling/views/payment.py
```diff
@@ -120,3 +124,89 @@ def post(self, request, *args, **kwargs):
...
+            raise Http404('Charge id not found')
+
+        try:
+            charge = stripe.Charge.retrieve(charge_id,
+                                            expand=['balance_transaction'])
+        except stripe.error.InvalidRequestError:
+            raise Http404('Charge not found on stripe')
+
+        try:
+            application_fee = stripe.ApplicationFee.all(charge=charge).data[0]
```
melinath: I think we should probably handle the refund even if there are no associated application fees?

chigby:
I need to go over this again to make sure it's doing all the right things, based on things I've learned about Stripe since.

brambling/models.py
```diff
@@ -1979,6 +1979,11 @@ class SavedReport(models.Model):
     querystring = models.TextField()
 
 
+class ProcessedStripeEvent(models.Model):
+    stripe_event_id = models.CharField(max_length=35)
+    created = models.DateTimeField(auto_now_add=True)
```
melinath: It seems like this model should have a couple more things, namely:

- whether this is the live or test API
- a unique constraint on the event id (and on whether it's live or test, if that's relevant)

brambling/views/payment.py
```diff
@@ -123,3 +126,92 @@ def post(self, request, *args, **kwargs):
...
+        try:
+            event = stripe.Event.retrieve(stripe_event_id)
+        except stripe.error.InvalidRequestError:
+            raise Http404('Event not found on stripe')
+
+        if event.type != 'charge.refunded':
+            return HttpResponse(status=200)
+
+        _, new_event = ProcessedStripeEvent.objects.get_or_create(
+            stripe_event_id=stripe_event_id)
```
melinath: If ProcessedStripeEvent needs to distinguish between live/test modes for uniqueness, that would need to be added to this query.

chigby:
In thinking about this some more, I felt like mingling the data for both test and live modes in the same table might actually be something we want to avoid entirely. Maybe this is me being paranoid and afraid of potential though not actual problems. But it's also not actually that hard to achieve here, since these tables are free-floating and lack relationships. So I went ahead and made the two separate and it seems to work fine.

melinath:
While I don't think that it's necessarily the worst to have two tables for this one thing, I have a couple of concerns.

This is different than how we handle test/live mode on all our other models. I would have a preference for us being consistent, which would mean keeping these as one model. I am tempted by the idea of splitting all our models in the future, since it would let us automatically enforce test/live data integrity at the database level. But I think that would make the models really hard to maintain and use - choosing whether to use LiveTransaction vs TestTransaction would be more annoying than choosing whether to pass live / test as an argument to filter / create. I'm also not convinced that the gain in database-level "validation" is significant over us just having better testing (which we would need anyway.)

So yeah... I'm leaning towards wanting these to be one model to be consistent with our other models. We might want to go with separate models later, we might not - but even if we did, it's unclear how long it would be until we could make that change, and we need to be able to deal with a consistent codebase in the meantime.

harrislapiroff:
We forgot to celebrate this pull request's birthday!

brambling/tests/functional/test_webhooks.py
```diff
@@ -144,3 +159,229 @@ def test_confirmation(self):
...
+    @vcr.use_cassette(os.path.join(VCR_DIR, 'stripe_webhook_refund_not_found.yaml'))
+    def test_stripe_refund_not_found(self, refund_retrieve):
+        refund_retrieve.side_effect = stripe.error.InvalidRequestError(
+            'not found', 1)
+
+        with self.assertRaises(Http404):
+            self.view(self.request)
+
+    @mock.patch('stripe.Charge.retrieve')
+    @vcr.use_cassette(os.path.join(VCR_DIR, 'stripe_webhook_charge_not_found.yaml'))
```
melinath: This test (and the others here) seem to use cassettes to mock a response from stripe even though they also use mock to patch the retrieval method to always error (which I think would mean that the cassette would never be used? Can the cassettes be removed?

brambling/tests/functional/test_webhooks.py
```diff
@@ -144,3 +159,229 @@ def test_confirmation(self):
...
+            charge,
+            event=self.event,
+            order=self.order,
+        )
+
+        self.refund = stripe_refund(self.order, self.event, charge.id, 100)
+
+        data = mock.Mock(object=mock.Mock(name='charge', id=charge.id))
+        self.mock_event = mock.Mock(
+            data=data, type='charge.refunded', livemode=False)
```
melinath: styleguide: should be on separate lines.

brambling/tests/functional/test_webhooks.py
```diff
@@ -144,3 +159,229 @@ def test_confirmation(self):
...
+        response1 = self.view(self.request)
+        response2 = self.view(self.request)
+
+        self.assertEqual(response1.status_code, 200)
+        self.assertEqual(response2.status_code, 200)
+        Transaction.objects.get(related_transaction=self.txn,
+                                transaction_type=Transaction.REFUND)
+
+    @mock.patch('stripe.Event.retrieve')
+    def test_events_should_be_processed_exactly_once_in_livemode(self, event_retrieve):
```
melinath: Since stripe_prep() isn't run again and self.txn isn't updated, this would seem to imply that we don't care if there's a mismatch between the event livemode and whether the Transaction is live or test.

Potential issue?

brambling/views/payment.py
```diff
@@ -123,3 +132,98 @@ def post(self, request, *args, **kwargs):
...
+        if event.type != 'charge.refunded':
+            return HttpResponse(status=200)
+
+        if event.livemode:
+            api_type = ProcessedStripeEvent.LIVE
+        else:
+            api_type = ProcessedStripeEvent.TEST
+
+        _, new_event = ProcessedStripeEvent.objects.get_or_create(
+            api_type=api_type, stripe_event_id=stripe_event_id)
```
melinath: styleguide: should be split onto separate lines

brambling/views/payment.py
```diff
@@ -123,3 +132,98 @@ def post(self, request, *args, **kwargs):
...
+
+        if not new_event:
+            return HttpResponse(status=200)
+
+        try:
+            charge_id = event.data.object.id
+            txn = Transaction.objects.get(remote_id=charge_id)
```
melinath: I think this needs to distinguish based on live vs test.

chigby:
I *think* this takes care of the main issue, which is that we weren't checking the Txn live/test mode. Which we are now doing :D

chigby:
Also took the opportunity to clean up some of the things you pointed out, along with some other things I noticed while poking around in there.

brambling/tests/functional/test_webhooks.py
melinath: self.txn is now updated.

PR #816, opened 2016-07-01
Shorten Register button on dashboard

nmorduch:
Also remove Facebook button from dashboard.

### DIFF
--- a/brambling/templates/brambling/_event_list_item.html
+++ b/brambling/templates/brambling/_event_list_item.html
@@ -50,8 +50,8 @@
 					</a>
 				{% endif %}
 			{% else %}
-				<a href="{{ event.get_absolute_url }}" class="margin-trailer-half btn btn-default">
-					<i class="fa fa-ticket"></i> Register <span class="hidden-xs hidden-sm hidden-md">for <strong>{{ event.name }}</strong>!</span>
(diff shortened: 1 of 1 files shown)


melinath:
@nmorduch why remove the facebook button?

nmorduch:
Doesn't need to be on homepage and takes up a ton of space. Better imo to have it on the shop page, since that's where the event description is. Having it on the homepage seems a bit like there's not a description on the shop page.

nmorduch:
Mehhh it's probably fine either way.

melinath:
I wonder - what about just getting rid of that line of buttons altogether? And making the name link to the registration as well?

nmorduch:
we never a/b tested, but buttons are helpful for people who don't think to click on things that don't scream link

melinath:
I think the website and register buttons look pretty lonely now:

screen shot 2017-02-10 at 11 40 41 am

Adding the FB url back in would _help_, but also... people seem to do fine interacting with sites like Eventbrite, which does not have a separate "register" button.

screen shot 2017-02-10 at 11 40 04 am

nmorduch:
I could be convinced.

Naomi I. Morduch Toubman
Little Weaver Web Collective
littleweaverweb.com

melinath:
Closing in favor of #834.

PR #834, opened 2017-02-10
Simplify dashboard

melinath:
An attempt at simplifying the dashboard. This would obviate #816. Changes are:

* Removes register/website/facebook/manage buttons from dashboard.
* Moves registration status onto event banner image (where manage button was)
* Makes banner image slightly taller so the whole thing actually displays
* Tightens margins in event metadata (name, date, location)
* Removes date card to cover up less of the banner image
* Removes verbose alerts below event image (since users can be expected to view a specific order to find those details)
* Linkifies event name.

screen shot 2017-02-10 at 12 36 24 pm

### DIFF
--- a/brambling/static/brambling/sass/modules/_event-listing.sass
+++ b/brambling/static/brambling/sass/modules/_event-listing.sass
@@ -1,6 +1,6 @@
 .event-listing-image
 	width: 100%
-	padding-top: 27%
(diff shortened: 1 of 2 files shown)


nmorduch:
I might put the genres about the time and place, but looks nice

melinath:
Oops, this was not reviewed.

PR #835, opened 2017-02-10
Simplify dashboard

melinath:
An attempt at simplifying the dashboard. This would obviate #816. Changes are:

* Removes register/website/facebook/manage buttons from dashboard.
* Moves registration status onto event banner image (where manage button was)
* Makes banner image slightly taller so the whole thing actually displays
* Tightens margins in event metadata (name, date, location)
* Removes date card to cover up less of the banner image
* Removes verbose alerts below event image (since users can be expected to view a specific order to find those details)
* Linkifies event name.

screen shot 2017-02-10 at 2 58 58 pm

See previous discussion in #834

### DIFF
--- a/brambling/static/brambling/sass/modules/_event-listing.sass
+++ b/brambling/static/brambling/sass/modules/_event-listing.sass
@@ -1,6 +1,10 @@
 .event-listing-image
 	width: 100%
-	padding-top: (300 / 940) * 100%
+	// 300 / 940 is the height/width ratio of the banner image.
+	// We use this technique to scale the size of the element
+	// properly while still being responsive and avoiding
+	// layout shifting as images load.
+	padding-top: percentage(300 / 940)
 	background-image: $dancers-event-banner-url
 	background-size: cover
 	background-position: center
--- a/brambling/static/brambling/sass/modules/_event-listing.sass
+++ b/brambling/static/brambling/sass/modules/_event-listing.sass
@@ -1,6 +1,6 @@
 .event-listing-image
 	width: 100%
-	padding-top: 27%
+	padding-top: (300 / 940) * 100%
 	background-image: $dancers-event-banner-url
 	background-size: cover
 	background-position: center
--- a/brambling/static/brambling/sass/modules/_event-listing.sass
+++ b/brambling/static/brambling/sass/modules/_event-listing.sass
@@ -9,32 +9,12 @@
 	box-sizing: border-box
 	border-radius: $border-radius-base
 
-.datecard
-	position: absolute
-	top: 6px
-	left: 6px
-	text-align: center
-	width:  50px
-	height: 50px
-	border-radius: $border-radius-base
-	float: left
-	color: white
-	box-shadow: inset 0 0 0 1px rgba(0,0,0,.2)
-	background-color: rgba($yellow-dark,0.9)
-
-.datecard-month
-	margin: 5px 0 0
-	text-transform: uppercase
-	font-size: floor(($font-size-base * 1.25))
-	font-weight: $headings-font-weight-normal
-
-.datecard-day
-	margin: 0
-	line-height: 1em
-	font-size: floor(($font-size-base * 1.5))
-
 .event-listing
 	& > .relative-right
 		right: 6px
 	& > .relative-top
 		top: 6px
+
+	& > h3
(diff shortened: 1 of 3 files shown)


nmorduch:
This is good, though I would prefer not to lose the manage button on the homepage. Maybe we can have both of them?

nmorduch:
Just changed to this:

image

Also, I was fond of those little date cards, but don't really care enough to argue about it.

brambling/static/brambling/sass/modules/_event-listing.sass
```diff
@@ -1,6 +1,6 @@
 .event-listing-image
 	width: 100%
-	padding-top: 27%
+	padding-top: (300 / 940) * 100%
```
emilyhorsman: Can we document where this calculation comes from? Also sass has a [`percentage`](http://sass-lang.com/documentation/Sass/Script/Functions.html#percentage-instance_method) function for what I think you're trying to do with the `* 100%` deal. :)

brambling/static/brambling/sass/modules/_event-listing.sass
```diff
@@ -9,32 +9,16 @@
...
-	line-height: 1em
-	font-size: floor(($font-size-base * 1.5))
-
 .event-listing
 	& > .relative-right
 		right: 6px
 	& > .relative-top
 		top: 6px
+
+	& > h3
```
emilyhorsman: The ampersand here is unnecessary. Do we have a style rule for being explicit about this?

brambling/static/brambling/sass/modules/_event-listing.sass
```diff
@@ -9,32 +9,16 @@
...
+		margin-top: 10px
+		margin-bottom: 5px
+
+	& > h5
```
emilyhorsman: (See prev comment about ampersand above)

brambling/static/brambling/sass/modules/_text.sass
```diff
@@ -70,6 +70,10 @@ h1, .h1
 
 //// Buttons
 
+// There was a weird line height thing going on with buttons.
```
emilyhorsman: Can you be more specific in documenting what this `line-height` declaration accomplishes?

brambling/templates/brambling/_event_list_item.html
```diff
@@ -1,94 +1,71 @@
...
+		{% if is_registered %}
+			{% if event.order and event.order.has_cart %}
+				<a href="{{ event.get_absolute_url }}" class="btn btn-default tipped" title="Complete your registration now!" data-placement="bottom">
+					<i class="fa fa-exclamation-circle"></i> Cart expires in
+
+					{% td_to_dict event.order.cart_expire_time|td_timeuntil as time_left %}
+
+					<strong data-countdown="timer">
+						{% if time_left.days %}<span data-countdown="days">{{ time_left.days }}</span> day{{ time_left.days|pluralize }}{% endif %}
+						{% if time_left.hours %}<span data-countdown="hours">{{ time_left.hours }}</span>:{% endif %}<span data-countdown="minutes">{% if time_left.minutes < 10 and time_left.hours %}0{% endif %}{{ time_left.minutes }}</span>:<span data-countdown="seconds">{% if time_left.seconds < 10 %}0{% endif %}{{ time_left.seconds }}</span>
```
emilyhorsman: Can we clean all this up a bit? It being on one smushed line makes this hard to parse and would make future amendments difficult.

brambling/static/brambling/sass/modules/_event-listing.sass
melinath: I'm not gonna change this because it matches the rest of the file and I don't think there's a compelling reason _not_ to do this...

brambling/static/brambling/sass/modules/_text.sass
melinath: AFAICT this just increases the line height of buttons slightly, but this was a @nmorduch change - pinging in case she wants to add anything.

brambling/templates/brambling/_event_list_item.html
melinath: It looks like it's probably crammed on one line because we can't allow whitespace between the things on this line.

PR #838, opened 2017-02-11
Fixed attendee notes updating

### LINKED ISSUE #831: Organizer notes broken
From an organizer:

> Got a 500 error when I clicked button &quot;save notes&quot; trying to add &quot;Notes&quot; in the notes box with the order items in it. (Adding the note in the &quot;notes&quot; box without the items in the order listed in it worked – though no confirmation was displayed and it sort of looks like nothing has happened. I can confirm it worked by viewing the order in the &quot;orders&#39; link on the left navigation bar)

> To follow up on previous submission – the problem is when I try to add notes to the &quot;Attendee&quot; notes box, it gives a 500 error. But when I add notes to the order it allows it. Please fix as this is really obnoxious if I need to add notes to an attendee in a multiple-attendee order and also means I have to track special orders in lots of different places. Thanks!

melinath:
Resolved #831. Fixed saving of per-attendee notes from organizer order detail page, and added tests proving it.

### DIFF
--- a/brambling/views/organizer.py
+++ b/brambling/views/organizer.py
@@ -1298,11 +1298,13 @@ class OrderDetailView(DetailView):
                         break
         self.transaction_refund_forms = [TransactionRefundForm(t)
                                          for t in self.order.transactions.all()]
+
+        # Transaction refund forms are not handled by this view so don't
+        # get included here.
+        forms = [self.notes_form] + self.attendee_forms
         if can_edit_event:
-            forms = [self.payment_form, self.notes_form, self.transaction_refund_forms]
-        else:
-            forms = [self.notes_form]
-        return forms + self.attendee_forms
+            forms.append(self.payment_form)
+        return forms
 
     def get_context_data(self, **kwargs):
         context = super(OrderDetailView, self).get_context_data(**kwargs)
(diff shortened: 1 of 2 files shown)


brambling/views/organizer.py
```diff
@@ -1298,11 +1298,13 @@ def get_forms(self):
...
+
+        # Transaction refund forms are not handled by this view so don't
+        # get included here.
+        forms = [self.notes_form] + self.attendee_forms
         if can_edit_event:
-            forms = [self.payment_form, self.notes_form, self.transaction_refund_forms]
-        else:
-            forms = [self.notes_form]
-        return forms + self.attendee_forms
+            forms.append(self.payment_form)
```
emilyhorsman: What about `self.transaction_refund_forms`? Are they not necessary anymore? If not, we should remove L1299.

melinath: Transaction refund forms still need to be generated - they're used by get_context_data - but they are not one of the forms that needs to be processed on POST.

Basically, `get_forms()` generates all the forms, whether or not they're supposed to be displayed. The return value just determines whether the form will be processed on POST.

melinath: This is arguably a silly way of doing this but redoing it would require rewriting the view for (at this point) little benefit IMO.

emilyhorsman: Gotcha. Alright, approving.

PR #846, opened 2017-05-15
766 check warning

### LINKED ISSUE #766: Add a warning if payment by check date is passed or blank
On the event Settings & Payment view:

screenshot 2016-02-23 17 04 06

If the date is blank or set to something in the past, then payment by check will not show up as an option to registrants. This makes sense, but it may not be immediately obvious to an organizer.

raq929:
Shows a warning if check postmark date has passed.
Closes #766 

screen shot 2017-05-15 at 6 18 01 pm

### DIFF
--- a/brambling/models.py
+++ b/brambling/models.py
@@ -2,7 +2,7 @@
 from __future__ import unicode_literals
 
 from collections import OrderedDict
-from datetime import timedelta
+from datetime import date as dtdate, timedelta
 from decimal import Decimal
 import itertools
 import json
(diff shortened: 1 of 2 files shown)


harrislapiroff:
UX seal of approval. :100:

brambling/models.py
```diff
@@ -2,7 +2,7 @@
 from __future__ import unicode_literals
 
 from collections import OrderedDict
-from datetime import timedelta
+from datetime import date as dtdate, timedelta
```
emilyhorsman: Non-blocking, but elsewhere in the repo and in other projects I don't think we've bothered to qualify the `datetime.date` import

PR #849, opened 2017-05-15
Fix 500 when Organizers save item notes

### LINKED ISSUE #831: Organizer notes broken

emilyhorsman:
Closes #831. I've added reproduction instructions there.

OrderDetailView#get_forms erroneously produced a nested list of forms,
instead of appending a constructed list to another one. As it turns out,
this did not actually have much to do with notes themselves—this was
merely the avenue to discover the issue.

### DIFF
--- a/brambling/views/organizer.py
+++ b/brambling/views/organizer.py
@@ -1299,7 +1299,7 @@ class OrderDetailView(DetailView):
         self.transaction_refund_forms = [TransactionRefundForm(t)
                                          for t in self.order.transactions.all()]
         if can_edit_event:
-            forms = [self.payment_form, self.notes_form, self.transaction_refund_forms]
+            forms = [self.payment_form, self.notes_form] + self.transaction_refund_forms
         else:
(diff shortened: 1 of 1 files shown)


emilyhorsman:
Oops. PR #838 already addressed this.

emilyhorsman:
Closing in favour of #838

PR #851, opened 2017-05-16
Fix unauthenticated transfer 500 and prohibits transfers while payment is pending

### LINKED ISSUE #840: Initiating a transfer while not logged in causes 500 error
From user:

> I have a ticket to attend <event>, and unfortunately, I will not be able to attend. I tried transferring my ticket to a friend though dancerfly, and each time I click the "Submit" transfer button, I get to a 500 error page.

It looks like there's a bug in our system that a) allows people without accounts to try to initiate transfers, and b) errors out if people without accounts try to initiate a transfer. The simplest fix for this is to require people to have a user account in order to transfer tickets.

### LINKED ISSUE #850: Prohibit transfers while payment is pending for an order
Right now you can transfer an item in an order to an email address before your payment for the order has completed.

emilyhorsman:
Closes #840 
Closes #850 

* Protected the POST route
* Informs the user and leads them to sign up
screen shot 2017-05-16 at 12 54 48 pm

screen shot 2017-05-16 at 12 55 07 pm

### DIFF
--- a/brambling/forms/orders.py
+++ b/brambling/forms/orders.py
@@ -531,7 +531,7 @@ class TransferForm(forms.Form):
             raise ValidationError("You can't transfer an item to yourself!")
         return email
 
-    def clean_boughtitem(self):
+    def clean_bought_item(self):
         bought_item = self.cleaned_data['bought_item']
         if TransferInvite.get_invites(bought_item).exists():
             raise ValidationError("A transfer has already been initiated for this item.")
--- a/brambling/forms/orders.py
+++ b/brambling/forms/orders.py
@@ -533,6 +533,8 @@ class TransferForm(forms.Form):
 
     def clean_bought_item(self):
         bought_item = self.cleaned_data['bought_item']
+        if not bought_item.can_transfer():
+            raise forms.ValidationError("This item is not eligible for transfer.")
         if TransferInvite.get_invites(bought_item).exists():
             raise ValidationError("A transfer has already been initiated for this item.")
         return bought_item
--- a/brambling/templates/brambling/event/order/_attendee_summary.html
+++ b/brambling/templates/brambling/event/order/_attendee_summary.html
@@ -25,7 +25,7 @@
 							({{ item.item_name }})
 							{% if event.transfers_allowed %}
 								{% url "brambling_event_order_transfer" event_slug=event.slug organization_slug=event.organization.slug as transfer_url %}
-								{% if item.status == item.BOUGHT %}
+								{% if item.can_transfer %}
 									<a href='{{ transfer_url }}?bought_item={{ item.pk }}' class='pull-right'>
 										<i class='fa fa-exchange'></i>
 										Transfer
--- a/brambling/templates/brambling/event/order/_attendee_summary.html
+++ b/brambling/templates/brambling/event/order/_attendee_summary.html
@@ -29,6 +29,9 @@
 									<a href='{{ transfer_url }}?bought_item={{ item.pk }}' class='pull-right'>
 										<i class='fa fa-exchange'></i>
 										Transfer
+										{% if not user.is_authenticated %}
+										(account required)
+										{% endif %}
 									</a>
 								{% endif %}
 							{% endif %}
--- a/brambling/templates/brambling/event/order/transfer.html
+++ b/brambling/templates/brambling/event/order/transfer.html
@@ -9,6 +9,14 @@
 		{{ block.super }}
 	{% endwith %}
 
+	{% if not user.is_authenticated %}
+	<div class='alert alert-danger'>
+		Sorry!
+		<b>You need to <a href="{% url 'brambling_signup' %}">create an account</a> to transfer an item.</b>
+		You'll be able to claim this order right away if you <a href="{% url 'brambling_signup' %}">create an account</a> with the same email address.
+	</div>
+	{% endif %}
+
 	<div class='row'>
 		<div class='col-xs-6 col-xs-offset-3'>
 			<p>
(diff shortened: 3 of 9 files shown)


emilyhorsman:
This breaks the POST request. Do not merge this.

emilyhorsman:
(This now no longer breaks the POST request but am still working on a functional test)

emilyhorsman:
This is ready for review again!

nmorduch:
I'm ok with this though the orange on red is really bad. i'm also planning on change what the link to transfer looks like, but i'll just make sure to go off of (or after) this branch

harrislapiroff:
We should just almost always use panel default tbh. Orange panels are goofy imo.

harrislapiroff:
Oh wait that's not the question for this PR, lol.

harrislapiroff:
Bootstrap comes with link classes for links inside alerts: http://getbootstrap.com/components/#alerts-links

brambling/forms/orders.py
```diff
@@ -531,8 +531,10 @@ def clean_email(self):
             raise ValidationError("You can't transfer an item to yourself!")
         return email
 
-    def clean_boughtitem(self):
+    def clean_bought_item(self):
```
melinath: ...lollll

brambling/templates/brambling/event/order/_attendee_summary.html
```diff
@@ -25,10 +25,13 @@
 							({{ item.item_name }})
 							{% if event.transfers_allowed %}
 								{% url "brambling_event_order_transfer" event_slug=event.slug organization_slug=event.organization.slug as transfer_url %}
-								{% if item.status == item.BOUGHT %}
+								{% if item.can_transfer %}
```
melinath: I would like to register a concern that `item.can_transfer` always triggers a query. For now this shouldn't be an issue because this is only being used once on this page, but if you _wanted_ to cache the property I wouldn't object ;-)

brambling/templates/brambling/event/order/transfer.html
```diff
@@ -9,6 +9,14 @@
 		{{ block.super }}
 	{% endwith %}
 
+	{% if not user.is_authenticated %}
+	<div class='alert alert-danger'>
+		Sorry!
+		<b>You need to <a href="{% url 'brambling_signup' %}?email={{ order.email }}">create an account</a> to transfer an item.</b>
+		You'll be able to claim this order right away if you <a href="{% url 'brambling_signup' %}?email={{ order.email }}">create an account</a> with the same email address.
```
melinath: I would either want this to not tell them that it needs to be the same email address, or to tell them here what email address they used on the order so they don't feel like they have to go look it up again.

brambling/templates/brambling/forms/signup.html
```diff
@@ -23,7 +23,7 @@
 <div class="row">
 	<div class="form-group col-xs-12">
 		<a href="#extraName" data-toggle="collapse" class="{% if masthead %}masthead-link {% endif %}pull-right more-name-options"><small><i class="fa fa-plus"></i> More Name Options</small></a>
-		{% formrow form.email %}
+		{% formrow form.email with value=request.GET.email %}
```
melinath: Why handle this here instead of adding request.GET.email to the forms initial data in python? I would rather see it that way.

brambling/tests/functional/test_order_views.py
```diff
@@ -87,3 +94,88 @@ def test_comped__sends_email(self):
...
+        self.factory = RequestFactory()
+
+    def set_up_view(self, orderer=None, is_confirmed=False):
+        organization = OrganizationFactory(check_payment_allowed=True)
+        event_owner = PersonFactory()
+        OrganizationMember.objects.create(
+            person=event_owner,
+            organization=organization,
+            role=OrganizationMember.OWNER,
+        )
```
melinath: Why do we need an event_owner? It doesn't _look_ like it's being used anywhere...

brambling/tests/functional/test_order_views.py
```diff
@@ -87,3 +94,88 @@ def test_comped__sends_email(self):
...
+            is_confirmed=True,
+        )
+        view.post(view.request)
+
+        self.assertEqual(len(mail.outbox), 1)
+        self.assertIn('offered an item transfer', mail.outbox[0].body)
+
+    def test_unauthenticated_transfer_fails(self):
+        """Only authenticated users should be allowed to transfer items,
+        despite unauthenticated visitors being alloewd to place orders.
```
melinath: typo

brambling/forms/orders.py
```diff
@@ -531,8 +531,10 @@ def clean_email(self):
...
         bought_item = self.cleaned_data['bought_item']
+        if not bought_item.can_transfer():
```
melinath: We should also in `__init__` be setting the queryset for this to just be items that can transfer, perhaps?

brambling/tests/functional/test_order_views.py
emilyhorsman: It could just directly be `person=PersonFactory()`

melinath: Why do we need an organization member at all?

brambling/forms/orders.py
emilyhorsman: Oh yeah good call

brambling/templates/brambling/event/order/_attendee_summary.html
emilyhorsman: Moved to #853

brambling/forms/orders.py
emilyhorsman: Moved to #853

Issue #705, opened 2015-11-13
Improve transfer UI


melinath:
Right now, transferring works but is fairly non-intuitive. It would be cool to fix that.

nmorduch:
## Tranferring an item:

- [ ] **Make the transfer link a button** in the order summary, to make it clear that it's not a statement describing the ticket
  - [ ] Possibly hide this in a settings dropdown if there are other settings we might like to make available. (e.g. changing to whom the item is assigned?) It's a little weird how prominent this option is.
  - [ ] Possibly make it say "Initiate transfer" so that it is extra clear that it is an action
- [ ] **Add panels for "Transfer Pending" and "Transferred"** below the "Attendees" panel in the order summary. All panels only appear if they have contents.
- [ ] **"Transfer Pending" panel**
  - [ ] Has attendee names as above (to make it clear which thing you would get back if you cancelled)
  - [ ] Has a "Cancel transfer" button in place of the transfer button
- [ ] **"Transferred" panel**
  - [ ] Does not show attendee names, because they are no longer relevant
  - [ ] Shows a list of items and two whom they were transferred, with no links

## Receiving a transfer
- [ ] **Add an alert** that says that you are currently working on receiving a transferred item
- [ ] **Show from whom a transfer was received** in the receipt, e.g. "Transfer from Naomi"

PR #858, opened 2017-05-19
Pending and Completed Transfer Summaries in Order Summary

### LINKED ISSUE #855: List pending transfers from the order summary page
List pending transfers with a Cancel action (see #598) in the same location where transfers are initiated.

emilyhorsman:
Closes #855 

screen shot 2017-05-19 at 2 33 15 pm

See #705 for checkboxes to check on merge.


brambling/utils/invites.py
```diff
@@ -109,7 +110,9 @@ def send(self):
     @classmethod
     def get_invites(cls, content=None):
         kwargs = {'kind': cls.slug}
-        if content is not None:
+        if type(content) is QuerySet:
```
melinath: `ininstance(content, QuerySet)` more pythonic

brambling/utils/invites.py
```diff
@@ -109,7 +110,9 @@ def send(self):
...
+            kwargs['content_id__in'] = [obj.pk for obj in content]
+        elif hasattr(content, 'pk'):
```
melinath: We should either have this be `content is not None` or we should add a blanket else to throw an error if the content doesn't match expectations.

brambling/templates/brambling/event/order/_completed_transfers_summary.html
```diff
@@ -0,0 +1,21 @@
...
+			Completed Transfers
+		</div>
+	</header>
+
+	<ul class="list-group">
+		{% for item in transferred_items %}
+		<li class="list-group-item">
+			{{ item.item_option_name }}
+			({{ item.item_name }})
+			accepted by
```
melinath: `transferred to` would be clearer, I think?

brambling/views/orders.py
```diff
@@ -823,6 +823,28 @@ def get_context_data(self, **kwargs):
...
+            for invite in invites
+        ]
+
+        # This is pretty awful and hopefully only used as a stopgap measure
+        # before Invites get refactored. Then again, this comment is one of
+        # those kinds that stick around for years.
+        context['transferred_items'] = BoughtItem.objects.filter(
+            transactions__related_transaction__transaction_type=Transaction.TRANSFER,
+            transactions__related_transaction__order=self.order,
+        )
```
melinath: We need tests to make sure that pending_transfers and transferred_items are being calculated properly. I do believe they are but we need tests.

brambling/models.py
```diff
@@ -1486,6 +1486,9 @@ def can_transfer(self):
             not self.transactions.filter(is_confirmed=False).exists()
         )
 
+    def get_transfer(self):
```
melinath: I would prefer not to add another single-purpose method that has side effects (additional queries).

brambling/views/orders.py
melinath: Re: removal of get_transfer, could we do something here similar to the pending transfers? I.e. do a prefetch of transactions related to the bought item, then return a dict of the item and the relevant transaction.

emilyhorsman: I've gone ahead and done this, but I think I disagree because the number of items is small (and thus the number of queries) and `BoughtItem#get_transfer` was more readable, IMO.

emilyhorsman:
@melinath This is ready for re-review.

brambling/static/brambling/sass/modules/_list-item.sass
```diff
@@ -0,0 +1,5 @@
+.list-item
+	display: flex
+	flex-direction: row
+	justify-content: space-between
+	align-items: center
```
melinath: I'm not sure if it's safe for us to use flexbox on the site? @nmorduch says it should be fine for non-essential purposes. Not sure whether this change counts as non-essential?

nmorduch: non-essential means that you can still use the site if flexbox fails

harrislapiroff: Also gonna make give an honorable mention to `display: table-row` and `display: table-cell` which have been able to do many of the flexy things we love with wide browser support for ages, though sometimes with wonky side effects. Not sure it fits in this context, but it looks like it might.

harrislapiroff: After some deliberation in Slack, we have determined that no one cares about IE9 (only 0.18% of traffic for the past 365 days) and everything else more or less supports flexbox. (Though we should double-check that we have a working autoprefixer setup?)

PR #859, opened 2017-05-19
Allow transfer cancellation

### LINKED ISSUE #598: Allow cancel of transfer
Right now, users can initiate a transfer, but there's no way for them to cancel it! Most of the time that shouldn't be a problem, but.

emilyhorsman:
Closes #598 

Must merge #858 before this one.

### DIFF
--- a/brambling/templates/brambling/event/order/_pending_transfers_summary.html
+++ b/brambling/templates/brambling/event/order/_pending_transfers_summary.html
@@ -18,8 +18,8 @@
 			</span>
 
 			<a
-				href="#"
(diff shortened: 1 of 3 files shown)


emilyhorsman:
@melinath I don't think linking an existing `DeleteView` requires tests

melinath:
@emilyhorsman I agree but you're also changing the behavior of the delete view, which _does_ need tests.

melinath:
More specifically you're changing what counts as having management permissions for a Transfer invite

PR #869, opened 2017-06-01
12 factor dev settings

harrislapiroff:
* Make `DATABASE_URL` and `SECRET_KEY` configurable environment variables.
* Clean up settings.py
* Allow for a local_settings.py file that provides per-installation overrides

This PR deliberately does not make any changes to production settings but it does suggestively posit that someday the dev settings might become the production settings.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ python:
 - '2.7'
 sudo: false
 env:
-- DATABASE_URL=sqlite:///db.sqlite3
 - DATABASE_URL=postgres://postgres@localhost:5432/brambling_test
 cache: pip
 install:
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -3,19 +3,19 @@ python:
 - '2.7'
 sudo: false
 env:
-- DB=sqlite3
-- DB=postgres
+- DATABASE_URL=sqlite:///db.sqlite3
+- DATABASE_URL=postgres://postgres@localhost:5432/brambling_test
```
melinath: TBH I would be fine with us only testing against postgres, since this is an application and not a reusable app, and we should just always be using postgres.

PR #874, opened 2017-06-15
Removed  attendees from table view

melinath:
We fucked up. The existing behavior was the intended behavior.


melinath:
The good news is this won't actually _break_ anything because attendees with transferred items won't ever exist. Hopefully.

PR #882, opened 2017-11-09
Switched stripe handling to use Stripe Elements

melinath:
For PCI compliance and to avoid having to fill out annoying forms every year, we need to switch to using Stripe Elements. This only affects client-side code.

To test this PR:

- Add payment env vars to your dancerfly activate file from 1password
- Create a demo user (email: demo@dancerfly.com)
- Create a demo event (org slug: `demo`; event slug: `demo-event`)
- Create one item for the demo event.
- Connect the demo org to LW stripe account at `/demo/edit/payment/`

- Verify that setting of publishable key is the same, code-wise.
- Verify that, using test API, you are able to save a credit card or get errors from `/billing/`
- Verify that, using test API, you are able to check out for an event both with and without saving the card.

### DIFF
--- a/brambling/templates/brambling/_stripe_js.html
+++ b/brambling/templates/brambling/_stripe_js.html
@@ -1,38 +1,20 @@
-<script type="text/javascript" src="https://js.stripe.com/v2/"></script>
 <script>
 	$(function(){
+		// External code must set stripe and card vars correctly.
 		$('#card-submit').on('click', function(e){
 			e.preventDefault();
 			$('.has-error').removeClass('has-error');
 			$('.help-block').remove();
 
-			Stripe.card.createToken({
-				number: $('#card-number').val(),
-				cvc: $('#card-cvc').val(),
-				exp_month: $('#card-expiry-month').val(),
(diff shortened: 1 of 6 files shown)


brambling/templates/brambling/_stripe_js.html
```diff
@@ -1,41 +1,23 @@
...
-						fieldId = '#card-cvc'
-					}
-					if (fieldId) {
-						var group = $(fieldId).parent();
-						group.addClass('has-error');
-						group.append('<p class="help-block">' + response.error.message + '</p>');
-					}
+			stripe.createToken(card).then(function(result) {
+				if (result.error) {
+					console.log(result.error);
```
raq929: I feel like we should remove console.logs in production.

melinath: I agree... I left it in because it was in the old version, but I was a little torn.

chigby:
When I attempt to pay for an event using a saved card, I get this message:

image

chigby:
When I attempt to pay with a non-saved credit card (entering a new one's information on the order summary page), I get this message:

image

PR #884, opened 2018-01-11
Reorganized API around endpoints instead of class types

melinath:
This should make working with the API simpler. Put it into a separate PR so that it can be evaluated separately for parity before changing things about the API.


harrislapiroff:
@melinath My review doesn't count because I don't have write access :p but I authorize you to give me write access and then merge this PR.

PR #889, opened 2018-03-16
Heroku

melinath:
This PR adjusts django-brambling to run on Heroku.


harrislapiroff:
I was able to successfully set up an event and get through the shop on the Heroku instance of the site, so mission critical stuff seems to be working. Will do a code review tomorrow.

Pipfile
```diff
@@ -0,0 +1,64 @@
...
+requests = "==2.10.0"
+dwolla = "==2.0.7"
+django-talkback = "==0.1.2"
+djangorestframework = "==3.3.0"
+openpyxl = "==2.2.6"
+jdcal = "==1.0.1"
+unicodecsv = "==0.14.1"
+markdown = "==2.6.5"
+bleach = "==1.4.2"
+"html5lib" = "==0.9999999"
```
harrislapiroff: jfc

brambling/payment/tests/test_dwolla.py
```diff
@@ -51,7 +52,7 @@
 
 
 class DwollaChargeTestCase(TestCase):
-
+    @skipUnless(os.environ.get('DWOLLA_TEST_APPLICATION_KEY'), 'dwolla test settings required')
```
harrislapiroff: Clarification: Is there something about the new setup that required you to add these skip decorators, or was this just to aid in local dev?

melinath: This was added because these tests require API keys, which we don't have a secure way to share with folks now that we're going 12-factor & 100% open-source. We can still keep the test API keys locally for our own use and on travis for CI, but we need to skip them if they aren't present.

Pipfile
melinath: yyyyyyyup

melinath:
@harrislapiroff updated the readme and responded to your questions!

brambling/payment/tests/test_dwolla.py
harrislapiroff: As long as they're still running in CI, I'm happy.

harrislapiroff:
Approved! I'll let you do the honors.

PR #907, opened 2018-07-18
File cleanup

harrislapiroff:
Removes these files that I'm fairly sure we no longer need:

* Python packaging files
* Tox file
* Fabric file


melinath:
We need the packaging files if we're going to use pip to install the repo editably, I think? But I guess we don't need to do that, since the code will be added to the path automatically by virtue of its location...

PR #922, opened 2019-01-05
[921] Removed RactiveShopView

### LINKED ISSUE #921: Remove RactiveShopView and supporting code
We aren't using RactiveShopView or any of its dependencies - see #647 - so we should remove it from the codebase to reduce technical debt and pave the way for a better shop replacement.

melinath:
This should be pretty straightforward - these files were only used by themselves and never actually got put into production, so we should be able to safely remove them.

Resolved #921


melinath:
Same thing happened to me on rebase... weird.

Issue #763, opened 2016-02-16
Apply discounts per-item


melinath:
Basically, instead of applying discounts to an entire cart, each time you entered a discount you would select one or more items in your cart to apply it to.

Right now, when a person who is volunteering registers and also buys the same pass for someone else, the discount is applied to both passes, which causes confusion and potential for errors.

We need to get the UI on this right...

melinath:
Current design for cart:

<img width="480" alt="screen shot 2019-03-06 at 10 05 37 pm" src="https://user-images.githubusercontent.com/299979/53935846-bdfaa600-405c-11e9-86e9-ca4f15d8fd81.png">

Problems with this design:

- Unclear which discounts are attached to which items
- Duplicated information in an attempt to make the above clear (but it doesn't)

Proposed design for cart:

![new-design](https://user-images.githubusercontent.com/299979/53937264-775b7a80-4061-11e9-9c0e-f8362b5752fb.png)

The first step will be to group the discounts with the items they're discounting, as shown here. The second step would be to add a "remove discount" link (here: trash can) that deletes a given discount.

PR #924, opened 2019-01-05
[763] remove orderdiscount model

melinath:
Removed OrderDiscount model and made it so that:

- Adding items to cart does not automatically add discounts to them.
- Adding discounts to the order adds them to any bought items that don't have them yet (instead of only adding them the first time the discount is applied)

This is the second phase of #763 and can be deployed standalone.


melinath:
Yeah we had a concept of the discount being added to the order (in addition to being added to each item individually) and it turns out it was a bad idea nobody wanted :joy:

harrislapiroff:
Oh I see. Even with this PR it still does that, but it doesn't add to past and future checkout processes. Is that correct?

melinath:
Yes, that's exactly correct :-)

PR #927, opened 2019-01-14
901 remove dwolla

### LINKED ISSUE #901: Discontinue dwolla support
After a brief conversation with @melinath we have a tentative agreement to discontinue Dwolla support. Dwolla's fee structure is attractive to organizers, but the maintenance burden has been unwieldy for us. Until we implement this, it's still open for discussion.

melinath:
Resolves #901. Removed dwolla support. This involved removing a number of views, some models and model fields, and references in about pages. I Ieft it in place as an option for transactions only, because we'll want to preserve information about past dwolla transactions (though there weren't ever many.) I did remove support for automatically linking to dwolla transactions using the remote_id field; if it becomes an issue we will still have the data and can find a new way to display it.


brambling/urls.py
```diff
@@ -76,8 +76,6 @@
     FinancesView,
 )
 from brambling.views.payment import (
-    DwollaConnectView,
-    DwollaWebhookView,
     StripeWebhookView,
 )
```
harrislapiroff: Wanna clean this import up while you're at it? 😇

brambling/views/user.py
```diff
@@ -173,14 +156,7 @@ def get_context_data(self, **kwargs):
             },
             'stripe_live_settings_valid': stripe_live_settings_valid(),
             'stripe_test_settings_valid': stripe_test_settings_valid(),
-            'dwolla_live_settings_valid': dwolla_live_settings_valid(),
-            'dwolla_test_settings_valid': dwolla_test_settings_valid(),
         })
```
harrislapiroff: Future thought: honestly weird that we give users access to test cards

brambling/forms/organizer.py
```diff
@@ -4,12 +4,21 @@
...
+from brambling.models import Event
+from brambling.models import Item
+from brambling.models import ItemImage
+from brambling.models import ItemOption
+from brambling.models import Order
+from brambling.models import Organization
+from brambling.models import OrganizationMember
+from brambling.models import SavedReport
+from brambling.models import Transaction
+from brambling.models import UNAMBIGUOUS_CHARS
```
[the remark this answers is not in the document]
melinath: Yes, it's easier to work with in the long run because there's never a transition between the two. The git history is much cleaner and it's only a little more verbose.

brambling/templates/brambling/user/billing.html
```diff
@@ -26,116 +26,93 @@
 	{{ block.super }}
 	<h2 class="sr-only">Billing settings</h2>
 
-	{% include "brambling/forms/_top_error_list.html" %}
+	{# {% if stripe_live_settings_valid or stripe_test_settings_valid %} #}
```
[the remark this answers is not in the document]
melinath: Originally this was not commented out; basically, it was set up to only display the stripe / dwolla settings if they were actually configured. I had commented it out for testing and forgot to re-comment it.

But looking at it again, I think we should just always display the stripe section; we already have messaging inside this section if the relevant stripe configuration isn't set up, and otherwise this screen is blank and confusing. So I'll remove the comment and de-indent the code inside it.

brambling/views/user.py
melinath: I think it's mostly for the demo user / event? But also - yes.

brambling/urls.py
melinath: done

PR #934, opened 2019-03-07
Removed zenaida hints usage

### LINKED ISSUE #918: Stop using zenaida hints
Currently we purport to use zenaida hints, but I'm not sure we ever actually implemented that. I don't _think_ it's a thing we care much about, in any case. If possible we should remove it as part of the project to remove zenaida as a dependency.

melinath:
Resolved #918. This was the only hint we had in place. I think we can get rid of it - I suspect that users will already be able to figure out how to use the site. And if we are concerned about the issues this hint seems to be trying to address, we should open tickets to handle them separately. Specifically:

- [ ] Redirect users back to events after creating accounts (which I would expect us to do already but if not it would be a good ticket to have)
- [ ] Remove user "dance preferences" as a feature. It's a weird UX and we don't have enough events at this time to try to be search-driven.

screen shot 2019-03-07 at 12 11 19 am


harrislapiroff:
Solid reasoning. I love it. I'm not concerned.

PR #938, opened 2019-03-08
Moved zenaida static files into brambling

### LINKED ISSUE #936: Move zenaida static files into django brambling
We currently use static files from zenaida - we should copy those over to folders in django-brambling instead.

melinath:
This depends on #935 and resolves #936. This work would otherwise conflict with #935 because both modify the base template significantly. Once #935 is merged, I'll change the base branch of this PR to master. Keeping it separate just makes the diff cleaner.


melinath:
Turns out this was just supposed to be the second commit on the branch; I got a little confused because of the force push history. Should be ready for review now!

PR #952, opened 2020-04-13
#951 Removed old event url shim

### LINKED ISSUE #951: Visiting missing org URL causes 500 error if multiple events have same slug as missing org
This has happened a couple of times recently. Essentially, back in the day, we didn't group events by organizations. `/<slug>/` would take you to an event. We put a shim in place to redirect people to an event by slug if no org existed with that slug; however, the workaround doesn't work if multiple events exist with the given URL; if multiple objects are returned, you instead get a 500 error.

Since the shim was added in 2015, I think we can probably remove it and just require people to have access to the full link.

melinath:
Resolved #951. Basically, we had a shim in place for old event URLs but the logic wasn't sufficient to handle the current load of events, and given it's 5 years old I think we can just get rid of it at this point.


harrislapiroff (approved):
💯

PR #957, opened 2020-04-27
956 upgrade django daguerre

### LINKED ISSUE #956: Upgrade to django-daguerre 2.3.1
This release supports django 1.11 and py2 and py3, so it would be a good choice to have in place for the 2->3 upgrade.

melinath:
Resolved #956. Upgraded django-daguerre to 2.3.1 and Pillow to 6.2.2. I believe that verifying that image resizing works for jpgs and pngs should suffice for verifying this change.


melinath:
Side note, since I did this in 2 commits, it seems like probably Pillow 6.2.2 works fine with the old daguerre version - but I don't think it hurts to upgrade them both at once.

